Issue #15, opened 2019-03-14
Allow OU targets in deployment_map.yml that do not (yet) contain accounts


thiezn:
Hi,

Would it be possible to add an OU that does not (yet) contain any accounts to be provided to the deployment_map.yml in the pipelines repository? This could be helpful when introducing new OUs to the AWS organisation. 

I noticed that the pipeline aws-deployment-framework-pipelines running in the deployment account gives the below error during the build phase:

```
========================== 22 passed in 0.31 seconds ===========================
 [Container] 2019/03/14 16:13:33 Running command python ./adf-build/generate_pipelines.py
Traceback (most recent call last):
 File "./adf-build/generate_pipelines.py", line 147, in <module>
 main()
 File "./adf-build/generate_pipelines.py", line 143, in main
 cloudformation.create_stack()
 File "/codebuild/output/src007560056/src/adf-build/shared/python/cloudformation.py", line 254, in create_stack
 create_change_set = self._create_change_set()
 File "/codebuild/output/src007560056/src/adf-build/shared/python/cloudformation.py", line 191, in _create_change_set
 ChangeSetType=self._get_change_set_type())
 File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 357, in _api_call
 return self._make_api_call(operation_name, kwargs)
 File "/usr/local/lib/python3.7/site-packages/botocore/client.py", line 661, in _make_api_call
 raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateChangeSet operation: [/Resources/Pipeline/Type/Stages/3/Actions] 'null' values are not allowed in templates
 [Container] 2019/03/14 16:13:37 Command did not exit successfully python ./adf-build/generate_pipelines.py exit status 1
[Container] 2019/03/14 16:13:37 Phase complete: BUILD Success: false
[Container] 2019/03/14 16:13:37 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: python ./adf-build/generate_pipelines.py. Reason: exit status 1

```

My deployment_map.yml file is similar to the below, where the /development OU does have an account, but the /production OU doesn't:

```
pipelines:
  - name: sample-vpc # The name of your pipeline (This will match the name of your repository)
    type: cc-cloudformation # The pipeline_type you wish to use for this pipeline
    params:
      - SourceAccountId: 111111111111
[...]

PR #16, opened 2019-03-14
Allow OU targets in deployment_map.yml that does not contain accounts

thiezn:
*Issue #, if available:*
#15 

*Description of changes:*
Changes to pipeline_repository to check if a target provided in deployment_map.yml actually contains accounts. If not it will skip trying to create a CloudFormation stack. This should prevent the failure as described in #15 

I wan't sure how to tackle writing a proper test for this. If you have some guidance or suggestions on this please let me know so I can update the pull request.

### DIFF
--- a/src/pipelines_repository/adf-build/generate_pipelines.py
+++ b/src/pipelines_repository/adf-build/generate_pipelines.py
@@ -113,9 +113,6 @@ def main():
                         raise Exception(
                             "Failed to return accounts for {0}".format(path))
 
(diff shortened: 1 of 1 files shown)


thiezn:
I'm cancelling the pull request for now as I'm seeing some unexpected behaviour that I want to make sure doesn't relate to the changes

thiezn:
Unexpected behaviour was due to something else. The changes in this pull request seems to solve the issue #15

thiezn:
see #15

Issue #27, opened 2019-04-03
Support for pipelines with build phase only


thiezn:
Hi,

We have a few lambda functions that we want to share between different accounts. To avoid code duplication we would like to have a single dedicated repository and pipeline for the lambda code. 

We would like to leverage ADF for the creation of the repository and pipeline but are only interested in the build stage of the pipeline. In the buildspec.xml we will package the code and upload to an S3 bucket.

I've tried using the cc-cloudformation type and providing an empty target list to the deployment_map.yml file similar to this:

```yaml
  - name: my-shared-lambda-function
    type: cc-cloudformation
    params:
      - SourceAccountId: 11223344
      - NotificationEndpoint: hello@moorspots.com
      - RestartExecutionOnUpdate: true
    targets: []
```

Unfortunately it fails during the template validation step:

```
cloudformation.validate_template()
 File "/codebuild/output/src223970246/src/adf-build/shared/python/cloudformation.py", line 130, in validate_template
 raise InvalidTemplateError(error.response['Error']['Message'])
errors.InvalidTemplateError: [/Resources/Pipeline/Type/ArtifactStores] 'null' values are not allowed in templates
```

*note* I did see another adf type called cc-s3 but haven't investigated yet if we can use this for our use case. However, I can imagine there are other use cases for this functionality not involving an S3 bucket.

Another solution might be to introduce a pipeline_type that allows you to define your own pipeline deployment stages. Not sure what that would look like but could give flexibility to someone wanting to leverage adf for anything they want to build.

Cheers,
Thiezn

PS. I realize you can leverage Lambda layers for code de-duplication but that only allows you to share modules. It won't allow you to share the full lambda itself for re-use in other templates.

bundyfx:
Closing as this is now resolved with #33

PR #31, opened 2019-04-13
Added cc-buildonly pipeline type

thiezn:
*Issue #27

*Description of changes:*
Added new cc-buildonly pipeline type template. When using this template the pipeline will only contain the build stage. The cc-buildonly.yml.j2 template is based on the cc-cloudformation.yml.j2 template.

### DIFF
--- a/src/pipelines_repository/adf-build/target.py
+++ b/src/pipelines_repository/adf-build/target.py
@@ -102,6 +102,9 @@ class Target():
         responses = self.organizations.dir_to_ou(self.path)
         self._create_response_object(responses)
 
+    def _target_is_region_only(self):
+        pass
+
     def fetch_accounts_for_target(self):
         if self.path == 'approval':
             return self._target_is_approval()
@@ -114,5 +117,8 @@ class Target():
 
         if (str(self.path)).startswith('/'):
             return self._target_is_ou_path()
+
+        if self.path is None:
+            return self._target_is_region_only()
 
         raise InvalidDeploymentMapError("Unknown defintion for target: {0}".format(self.path))
--- a/src/pipelines_repository/adf-build/tests/stubs/stub_deployment_map.yml
+++ b/src/pipelines_repository/adf-build/tests/stubs/stub_deployment_map.yml
@@ -27,3 +27,12 @@ pipelines:
       - NotificationEndpoint: my_email@email.com
     targets:
       - 123456789
+
+  - name: buildonly
+    type: cc-buildonly
+    params:
+      - SourceAccountId: 11111233321
+      - NotificationEndpoint: my_email@email.com
+    targets:
+      - regions: eu-west-1
+
--- a/src/pipelines_repository/adf-build/deployment_map.py
+++ b/src/pipelines_repository/adf-build/deployment_map.py
@@ -63,7 +63,7 @@ class DeploymentMap:
                 for target in pipeline["targets"]:
                     if isinstance(target, dict):
                         # Prescriptive information on the error should be raised
-                        assert target["regions"] and target["path"]
+                        assert target["regions"]
         except KeyError:
             raise InvalidDeploymentMapError(
                 "Deployment Map target or regions specification is invalid"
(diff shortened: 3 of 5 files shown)


bundyfx:
Hi @thiezn 

Thanks for the PR!

How does one use this type in their *deployment_map.yml*?

What I feel makes sense *(and to keep things standardized)* to me is something like:

```
  - name: my-build
    type: cc-buildonly
    params:
      - SourceAccountId: 12345678910
```

However in ADF's current form **targets** is a required key in the deployment map. We could make targets optional of course, thus allowing this above syntax to work as expected. Interested to hear your thoughts on this.

Cheers,
Flynn

src/pipelines_repository/adf-build/tests/stubs/stub_deployment_map.yml
```diff
@@ -27,3 +27,12 @@ pipelines:
...
     targets:
       - 123456789
+
+  - name: buildonly
+    type: cc-buildonly
+    params:
+      - SourceAccountId: 11111233321
+      - NotificationEndpoint: my_email@email.com
+    targets:
+      - regions: eu-west-1
```
bundyfx: Currently we don't have the possibility to have the "build" phase in a different region other than the deployment region. I feel this would make more sense to just remove the "targets" key.

src/pipelines_repository/adf-build/target.py
```diff
@@ -102,6 +102,9 @@ def _target_is_ou_path(self):
         responses = self.organizations.dir_to_ou(self.path)
         self._create_response_object(responses)
 
+    def _target_is_region_only(self):
```
bundyfx: Im guessing you are wanting to check something here?

src/pipelines_repository/adf-build/target.py
```diff
@@ -115,4 +118,7 @@ def fetch_accounts_for_target(self):
         if (str(self.path)).startswith('/'):
             return self._target_is_ou_path()
 
+        if self.path is None:
```
bundyfx: As per comment in stub_deployment_map - we don't have a j2 template to do cross region builds, nor has it been tested yet.

thiezn:
Hi @bundyfx 

Thats a good point, at the moment I was using a dummy target in the deployment map. I had a look at the code and the difficulty with your proposed deployment_map is that the build stage does require an output artifactstore. This is currently dependent on the region selected in the targets section. 

Perhaps a solution to this would be making the path / account step optional in the deployment map:

```
 - name: my-build
    type: cc-buildonly
    params:
      - SourceAccountId: 12345678910
    targets:
       - regions: [ eu-west-1, eu-central-1 ]
```
The added (small) benefit is that you can leverage the ADF s3 resources in your buildstep for your own needs

I've updated my pull request with some changes that would allow you to omit the path in the deployment map. Let me know what you think, it might not be a direction you want to take this...

Cheers,
Thiezn

src/pipelines_repository/adf-build/tests/stubs/stub_deployment_map.yml
thiezn: Yes, i agree that would make the most sense but the problem is that the build stage wants write its output to a ArtifactStore thats currently being generated based on the regions provided in the target.

src/pipelines_repository/adf-build/target.py
thiezn: This seemed to be the easiest way to get a targets section without a path defined.

bundyfx:
Hi @thiezn - I cannot amend to your PR so I created a new PR with your commit in it -> #33 

I tested this out and it works as I expect, let me know what you think and we can merge this in. Cheers and thanks for the PR! 🎉

thiezn:
Ha great @bundyfx !

I missed the 'deployment_account_region' variable in the jinja2 templates. This makes a lot more sense, thanks for the effort.

I'll leave my PR until you've merged yours, not sure if the changes get lost when I cancel mine.

Cheers,
Mathijs

bundyfx:
Great! Merging the other one now with your commit in it. Thanks again!

PR #46, opened 2019-05-01
0.3.1

### LINKED ISSUE #43: Idea: Allow configuring custom stage names
E.g. like this:

```
- path: /banking/testing
  regions: eu-central-1
  name: my-stage-name
```

### LINKED ISSUE #44: Idea: allow use of all CloudFormation update types (e.g. also "Create or update a stack")
Possible types:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/continuous-delivery-codepipeline-action-reference.html

bundyfx:
*Issue #, if available:*

Resolves #43 
Resolves #44 

*Description of changes:*

#43 - You can now have custom names for stages which are optional:

```
    targets:  
      - path: 1111111111
        name: my-step
      - path: approval
        name: manager_plz_approve
      - path: 99999999999
        name: my-other-step
```

This is backwards compatible with the older syntax and will default to the names 'deployment-stage-n' if not specified. This makes it much easier to attach IAM policies that are tied to approval stages in CodePipeline since the names can now be static/predictable - see [more here](https://docs.aws.amazon.com/codepipeline/latest/userguide/approvals-iam-permissions.html) on that.

#44 - You can now specify an action for a pipeline when using CloudFormation *(only)* which uses one of the CloudFormation action types as define [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/continuous-delivery-codepipeline-action-reference.html#w2ab1c13c13b9).

```
  - name: sample-node-app
    type: cc-cloudformation
    action: replace_on_failure # See link above for other examples here (eg create_update) 
    params:
      - SourceAccountId: 1111111111
      - Image: aws/codebuild/docker:18.09.0
    targets:
      - path: 111111111111
      - approval
      - 99999999999
```

### Other changes
* Updating sample guide for clarity + examples
* Giving definition and example of shorthand targets syntax vs path, regions, name style.
* Updating j2 templates to support new step name structure
* Updating tests to support new step naming structure
* Update SCP logging to output OU path name not OU id
* Cleaning up leftover imports/exports logic in j2 templates *(was not used)*
* Slight refactoring of using getters/setters where possible *(more to come on this refactoring)*


oozeorb:
Hi @bundyfx , can you also make an example with the SAM framework?
greetz

Issue #88, opened 2019-07-01
Create CodeCommit repo from deployment_map entry


triha74:
I would really like to simplify creation of a new repo to make it "one click to get repo and pipe".

By adding a "create-repo: true" attribute ADF should be able to create a code commit repo by launching a CF in the SourceAccount. The CF could be a template provided by the pipeline generator repo so that users can extend it.
```
pipelines:
  - name: my-sample-app
    create-repo: true
    type: cc-customdeploy
    params:
      - SourceAccountId: 9876543210
    targets:
      - path: 12345667899
```

bundyfx:
Agreed I think the repo should be automatically created if using CodeCommit as a source, we have the account Id and we know the repo name so this makes sense. I think this would be best as the default if using CodeCommit, I cannot see a scenario when you would not want the repo created. Should we bother with the 'create-repo' key?

triha74:
It might be that users want to create the repo them selfs using CF and including event lambdas and stuff.

bundyfx:
Ok, I like the idea. Maybe we should make it the default then, and it can be opt out of with:

```
create-repo: false
```

thiezn:
+1

We currently use a seperate ADF managed pipeline that just churns out CodeCommit repositories including some IAM Roles that prevents pushing directly to master. Recently AWS CodeCommit allows you to provide initial content (e.g. README.md and .gitignore templates) by pointing to a zip file on an s3 bucket.

Would be great if this can be made possible through ADF as well. We could add another parameter pointing to the zip file:

````
pipelines:
  - name: my-sample-app
    create-repo: true
    initial-repo-content: arn::s3:mybucket/initialrepocontent.zip
```

thiezn:
Something else, shouldn't the create-repo parameter fit better in the params section like so:

```yaml
pipelines:
  - name: my-sample-app
    type: cc-cloudformation
    params:
     - create-repo: true
     - initial-repo-content: arn::s3:mybucket/initialrepocontent.zip
```

triha74:
After reading this thread I actually think you either want to have ADF create your repos or you dont. I dont really see that you want it to manage some of them.

So I actually think a global `autogenerate-code-repositories: true | false` is what we would want.

thiezn:
Perhaps, I don’t think the deployment map supports any global config parameters at the moment.


I do feel you want to be able to have different initial commit for repos created with ADF so that should not be a global param.

PR #93, opened 2019-07-05
Implemented support for ADF to create code commit repositories based …

triha74:
*Issue #88, :*

*Description of changes:* Added support to create code commit repos automatically from the config in deploy_maps.

A cloudformation stack is launched `adf-codecommit-${app}` to the specified source account if there is no repo in that account.

If there is a repo in that account and the `adf-codecommit-${app}` does exist the stack is updated. 

This way ADF only manages the code commit repo it was not created before adding the application to the deployment_map.


triha74:
I added opt in opt out through a common config file for the pipeline generation called config.yml.

PR #117, opened 2019-08-01
Fixing Github Example in User Guide

javydekoning:
**Issue #, if available:** N/A

**Description of changes:** Fixed the Github example, the previous one listed a `SourceAccountId` which only applies to CodeCommit. It was also missing param's required for Github.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -41,7 +41,6 @@ pipelines:
     type: github-cloudformation
     action: replace_on_failure
     params:
-      - ProjectName: "vpc-example" # Name of the Github repository
       - Owner: "githubrepowner" # Repository owner user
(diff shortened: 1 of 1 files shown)


docs/user-guide.md
```diff
@@ -36,14 +36,18 @@ pipelines:
...
         regions: eu-central-1
 
-  - name: vpc
+  # Github source example
+  - name: vpc-example # Pipeline name MUST be equal to Github repository name.
     type: github-cloudformation
     action: replace_on_failure
     params:
-      - SourceAccountId: 8888877777777
+      - ProjectName: "vpc-example" # Name of the Github repository
```
bundyfx: This parameter is not mandatory, ADF itself generates the value of this parameter and passes it in at Pipeline creation time, it will just uses the project name *(aka the pipeline name)*.

Issue #126, opened 2019-08-07
Multi-regional pipeline with 'contains_transform: true’ incorrectly reference template.yml


at7925:
Multi-regional pipeline with 'contains_transform’ incorrectly reference template.yml in TemplatePath:

Example:
_deployment_maps/defect.yml_
```
---
pipelines:
  - name: test-invalid
    type: cc-cloudformation
    regions: [ us-east-1, eu-west-1 ]
    contains_transform: true
    params:
      - SourceAccountId: 112233445566
    targets:
      - /deployment
```

Checking TemplatePath in pipeline:
```
aws codepipeline get-pipeline --name adf-pipeline-test-invalid | grep -B1 TemplatePath

          "TemplateConfiguration": "test-invalid-build::params/Deploymentaccount_eu-west-1.json",
          "TemplatePath": "test-invalid-build::template_.yml"
--
          "TemplateConfiguration": "test-invalid-build::params/Deploymentaccount_us-east-1.json",
           "TemplatePath": "test-invalid-build::template_.yml"
```
**Expected result:** template_us-east-1.yml and template_eu-west-1.yml accordingly with regions

**Actual result:**  invalid reference to template file: template_.yml

_Note:_ template will be referenced correctly if regions are defined on target level.

bundyfx:
Ah yes nice one, this is just a slight bug in the J2 when using contains_transform, I will make a patch release for this shortly unless you would like to make a PR within the J2 templates themselves.

bundyfx:
Resolved in 1.2.6! Thank you.

PR #129, opened 2019-08-13
fixing bug in the J2 when using contains_transform and top_level_regions (#126)

at7925:
*Issue #, if available:* 126

*Description of changes:*
fixing bug in the J2 when using contains_transform and top_level_regions


bundyfx:
Thank you!

Issue #128, opened 2019-08-13
CodePipeline limits 


azamin:
Hello, 

I am seriously worried about the CodePipeline limits. 
Let please somebody check my math and give me a hint how can I configure ADF to manage the baseline in 80 production accounts each of which operate in multiple (different) regions.

A pipeline can have maximum of 10 stages and 50 actions in a stage, this means theoretical maximum of 500 actions.
Two of the stages are always reserved for CodeCommit and CodeBuild, so it's actually 400 actions.
Each regional deployment needs 2 actions (replace->execute), so we go down to 200 account/regions per pipeline.

Because of the hard limit of 50 actions per stage (as per AWS Support), we can have maximum of 25 account/regions per stage. 
This can be something between one account with 25 regions or 25 accounts with one region.
A stage in ADF represents also an AWS Organizaton Unit, which means also maximum of 25 accounts per OU.

The only workaround that I found is to use multiple pipelines to the same repository, but we still have the limit to 25 accounts per region, so we have to separate them in OU buckets on the Organization 
like this:
regions: eu-west-1
/customers1/
/customers2/
..
regions: eu-central-1
/customers1/
/customers2/

Looks ugly and because of limit the 200 regional deployments per pipeline, we can have maximum of 8 OU buckets with maximum of 25 accounts per bucket.
Please let somebody tell me that I am wrong here, because I don't see how these 80 accounts will scale to 300 so that we can move our SaaS to AWS.  

Thank you

elmoenco:
I've literally came across this exact problem 10 minutes ago, moving a bunch (more than 25) of accounts into an OU. Extremely disappointing.

azamin:
@elmoenco, for me it was only 3 working days and their support employee answered:
"I have an update from the service team and they informed that the maximum actions in a stage that we support is 50, which is the hard limit for our system. "

Good luck with your case and please let me know if you got it increased so that escalate to our account manager.

elmoenco:
Thanks!

This is very bad. 
I can scale as far as I want with tools like concourse. This requires an abstracted approach of pipelines stages and jobs, all because of CodePipeline limitations.

azamin:
i love ADF and want to continue using it, so I decided for this workaround: 
Please let me know if there is better way

```
# aws-deployment-framework-pipelines/deployment_maps/baseline-guardduty.yml
pipelines:
 - name: baseline-guardduty
   type: cc-cloudformation
   contains_transform: true # because it's SAM
[...]

PR #137, opened 2019-08-19
showing and explaining the CodePipeline limits to early adopters

azamin:
issue #128 

It's important that early adopters know and understand the CodePipeline limits.
Let's be honest and add URL to the FAQ document on our main README file and also have one line explanation what does the limit means for the most common ADF pipeline type.

------


azamin:
changing the action mode increases the default limit from 200 to 400 accounts per region, which seems OK for me. Thanks

Issue #135, opened 2019-08-18
ThrottlingException - on adf-deployment-framework


amimarkels:
I assume that because we have more then 40 pipeline and because they run in different threads we get a "ThrottlingException - (reached max retries: 4): Rate exceeded" error **a lot** on the the adf-deployment-framework pipeline. 
trying to find a solution we came with boto allowing to increase the number of retries (https://botocore.amazonaws.com/v1/documentation/api/latest/reference/config.html), but we are opened minded for a more elegant solution.

amimarkels:
After further investigation it is even a bigger issue. 
We have 50 pipelines. 
In each run of the adf-deployment-framework at least one stack is failing. 
The reason is that in `get_stack_status` the describe_stacks call fails due to throttling issue. Due to this it returns a value of 'None' and when trying to create it, it is failing on stack exist:

botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateChangeSet operation: Stack [adf-pipeline-network-templates-xxx-us-east-1] already exists and cannot be created again with the changeSet [adf-pipeline-network-templates-xxx-us-east-1].

bundyfx:
Thanks @amimarkels 

Ok, I try replicate this myself and see if we can get a fix prioritized for this one. The stack [adf-pipeline-network-templates-xxx-us-east-1] most likely has a failed change set on it for a different reason, you can navigate to the CloudFormation failed change set in the console and see some related error information about it, ADF creates the same named change set each time it runs so it cannot create one if there is a failed change set already there.

I'll try replicate this now, we might need to change the config for retries as you suggested and also increase the frequency in which we check the stack status as part of the waiter.

at7925:
Lately experiencing the same *botocore.exceptions* :
```
botocore.exceptions.ClientError: An error occurred (ThrottlingException) when calling the PutParameter operation (reached max retries: 4): Rate exceeded 
```
and
```
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateChangeSet operation: Stack [adf-pipeline-stackname] already exists and cannot be created again with the changeSet
```
we have 60+ pipelines and 30+ accounts

bundyfx:
> Lately experiencing the same _botocore.exceptions_ :
> 
> ```
> botocore.exceptions.ClientError: An error occurred (ThrottlingException) 
> when calling the PutParameter operation (reached max retries: 4): Rate exceeded 
> ```
> 
> and
> 
> ```
[...]

PR #141, opened 2019-08-21
v1.2.8

bundyfx:
*Issue #, if available:*

Aims at solving #135 

*Description of changes:*
- Adds in retries (10) for SSM and CFN config.
- Adds in random delay for CloudFormation waiter to reduce the DescribeStack API Calls.


amimarkels:
@bundyfx We have implemented this fix today and it looks like it solved the problem.

PR #156, opened 2019-10-16
Fix to replace spaces and other unaccepted chars from CFn stack names

sbkok:
With this fix it replaces all unaccepted chars from the CloudFormation
(CFn) stack name with a dash.

For example, when the OU name would be 'IT Department', it would fail
setting up the base ADF resources as it would generate the following
stack name: 'adf-global-base-IT Department', which includes a space that
is not supported by CloudFormation.

With this fix, it will replace the space with a dash, the new name would
become: 'adf-global-base-IT-Department'. Any other unsupported character
is treated the same.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/test_cloudformation.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/test_cloudformation.py
@@ -6,10 +6,11 @@
 import os
 import boto3
 from pytest import fixture
+from unittest.mock import patch
 from stubs import stub_cloudformation
 from mock import Mock
 
-from cloudformation import CloudFormation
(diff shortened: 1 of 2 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/test_cloudformation.py
```diff
@@ -6,10 +6,11 @@
 import os
 import boto3
 from pytest import fixture
+from unittest.mock import patch
```
cattz: Is this import needed?

sbkok: Good catch! Thanks for that, I will remove it.

PR #161, opened 2019-10-18
Fix to skip deletion of deployment_maps dir and other json/yaml files

sbkok:
**Issue:**

When upgrading ADF, it deletes the old files that have been removed
between versions. However, the way it performs this test contained an
issue where it deleted the `deployment_maps` directory and other json/yaml
files that were not specifically whitelisted to be kept.

**Fix:**

Checks the files that would be scheduled for deletion against a regular
expression. When the file ends with a `json`, `yaml`, `yml` file extension
it should be kept in the repo as is.

Unit tests have been added to ensure that it would work as intended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,14 @@
 init:
 	pip install -r requirements.txt
+	pip install -r src/lambda_codebase/initial_commit/requirements.txt
 
 test:
 	# Run unit tests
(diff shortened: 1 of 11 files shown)


sbkok:
I've fixed the integration issue that was blocking this PR. It was missing
the default AWS region environment variable.

Additionally, I have ensured the fix would be applied in both
`initial_commit.py` files that exist in this repository.
I have updated the two files such that they are even more alike,
i.e. code layout/white space fixes. This makes it easier to compare them.

We would need to get this deduplicated later on.

PR #163, opened 2019-10-21
Tag based deployment targeting.

triha74:
*Description of changes:* Added the ability to use account tags in the target section of the pipeline definition. The schema as been updated to support "tags: [(string, string)]" as target.

Example:
```
     targets:
          - tags:                                 #The security teams staging
              team: security
              environment: staging 
            name: security-qa
          - approve
          - tags:                                 #All QA environments staging
              environment: qa 
            name: all-qa
          - approve
          - tags:                                 #All Prod environments staging
              environment: prod 
            name: all-prods
```
```
     targets:
          - tags:                                 #My teams dev account
              team: triha-dev-team
              environment: staging 
            name: team-test-env
          - tags:                                 #The service qa account
              service: ecom
              environment: qa 
            name: ecom-qa
          - approve
          - tags:                                 #The service prod account
              service: ecom
              environment: prod 
            name: ecom-prods
```
This allows us to use Organizations to model SCP policy groups and Tags to assign organizational/architectural attributes to the meta data of an account. This way we can deploy SCPs using Org paths and Applications using account meta data.


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/generate_pipeline_inputs.py
```diff
@@ -122,14 +127,21 @@ def worker_thread(p, organizations, auto_create_repositories, deployment_map, pa
     regions = []
     for target in p.get('targets', []):
         target_structure = TargetStructure(target)
+
         for step in target_structure.target:
+            regions = step.get(
+                'regions', p.get(
+                    'regions', DEPLOYMENT_ACCOUNT_REGION))
```
sbkok: Good fix!

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -22,9 +22,14 @@ class Organizations: # pylint: disable=R0904
...
 
     def __init__(self, role, account_id=None):
+        self.role = role
         self.client = role.client(
             'organizations',
             config=Organizations._config)
+        self.tags_client = role.client(
+            'resourcegroupstaggingapi',
+            region_name='us-east-1',
+            config=Organizations._config)
```
sbkok: Would be great to build this client from inside the Organizations class, so we don't need to access private variables like `_config`.
Alternatively we can change the config to be public as well, but I would prefer adding it to the class instead.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/target.py
```diff
@@ -95,6 +95,14 @@ def _target_is_account_id(self):
         ).get('Account')
         self._create_response_object([responses])
 
+    def _target_is_tags(self):
+        responses = self.organizations.get_account_ids_for_tags(self.path)
```
sbkok: Could you rename `responses` to `account_ids`?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/schema_validation.py
```diff
@@ -275,8 +275,11 @@
     int
 )]
 
+# Pipeline Params
+
 TARGET_SCHEMA = {
     Optional("path"): Or(str, int, TARGET_LIST_SCHEMA),
+    Optional("tags"): {str: object},
```
sbkok: Currently, the schema would allow empty strings as keys as well.
The value of the tags should be string as well right?

I think we can replace the definition to:
```python
Optional("tags"): { And(str, Regex(r"\\A.{1,128}\\Z")): And(str, Regex(r"\\A.{0,256}\\Z")) }
```
According to the [Tag Restrictions](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions) it can be any Unicode char for the key, as long as it is less than 128 chars long.
For the value, it can be any Unicode string up to 256 chars long.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/target.py
triha74: In the entire class the responses from boto3 calls are called responses. I think being consequent is better. The entire class could be refactored but I dont want to add functionality and refactor at the same time unless a refactoring is needed in order to add the functionality in a decent way.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/schema_validation.py
triha74: Ill add this.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
triha74: This is done exactly like the other client uses in the same class.

sbkok: Alright, I think we should get those changed too in that case. But having that in a separate PR would make sense I guess.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/target.py
sbkok: I would propose getting the boto3 calls renamed in that case too, but that would be a separate PR.

In this case we are not calling a boto3 function though.
You are invoking this [function (.../shared/python/organizations.py#L137-R153)](https://github.com/awslabs/aws-deployment-framework/pull/163/files#diff-f47c805f0dd7e769092ac78c5b434ed4R137-R153), or am I mistaken?

PR #165, opened 2019-10-23
Fixes uniqueness of deployment CodeBuild project names

sbkok:
**Issue:**

When using CodeBuild as a deployment target, the name that would be
generated for that stage would be structured as:

```
{target_aws_account_name}-stage-{index + 1}
```

Where the target AWS account is the deployment account for all
pipelines. In other words, if you try to recreate a similar setup where
the deployment stage that uses CodeBuild would be at the same stage
index, it will not be able to deploy the pipeline as the name is already
in use.

**Fix:**

Instead of using the target account name, it will use the pipeline name.

Refactored the name reference part in the same file as well.


sbkok:
Closing this PR, as #169 fixes the core issue of the naming of CodeBuild projects when used as deployment steps. I will open a separate PR for the other changes in this PR and rebase those on top of #169.

PR #176, opened 2019-10-31
CDK examples and Assume Role helper

triha74:
*Issue #, if available:*

*Description of changes:* Adding CDK Examples and an Assume Role helper.


at7925:
If this script can be called from ordinary buildspec and assume any deployment role in any account then we have a **BIG security issue in ADF**, see:
```
export TARGET_ACCOUNT_ID=$(aws ssm get-parameter --name master_account_id | jq -r '.Parameter.Value')
source adf-build/helpers/assume_role.sh adf-cloudformation-deployment-role
# user to be created in Master account or any developer can read any S3 bucket in any account (hello CapitalOne)
aws iam create-user --user-name hacker 
```
Obviously, this is not a problem of the helper script, perhaps default CodeBuild Role should be less powerful

bundyfx:
Im going to close this one and open a more specific issue around assume role access for the codebuild role.

Issue #188, opened 2019-11-18
Pipeline is not triggered by CodeCommit repositories changes if pipeline uses branch different from master


at7925:
Pipeline is not triggered by CodeCommit repositories changes if pipeline uses branch different from master. 
It looks like a Cloudwatch event for triggering pipeline on repository changes always uses master branch instead of explicitly defined default_providers:source:branch

bundyfx:
Thanks @at7925 - Indeed users should be able to use other branches than master.

Also thanks for the PR. One thing that is missing in the PR is the CloudWatch rule that gets created on the source account to trigger the pipeline. The rule that is created defines which branch should trigger the event in order to trigger the pipeline (current hardcoded "master"). We will need a way to pass the value (probably parameter store) from the map and into the source account so that we can define the cloudwatch rule correctly. Ill have a think about this one also but your PR mostly resolves the issue :)

PR #193, opened 2019-11-20
use default_providers:source:properties:branch when defined instead master

at7925:
*Issue #, if available:*
#188 Pipeline is not triggered by CodeCommit repositories changes if pipeline uses branch different from master
*Description of changes:*
use **default_providers:source:properties:branch** when defined instead of hard-coded master


bundyfx:
This is now part of 3.0, with some slight additions! Thank you!

PR #200, opened 2019-12-12
Document using OUs as deployment targets

cristim:
Apparently the use case for deploying to an entire OU wasn't documented yet.

*Issue #, if available:* N/A

*Description of changes:*

Documentation update for the use case of using OUs as deployment targets.


docs/admin-guide.md
```diff
@@ -313,6 +314,28 @@ pipelines:
...
+    default_providers:
+     source:
+        provider: codecommit
+        properties:
+          account_id: 11111111111111 # <-- This AWS account that hosts the source of the CloudFormation payload deployed by this pipeline.
+    targets:
+      deploy:
+        provider: cloudformation
+    targets:
+      - path: ou-d34d-b33f # <-- This is the OU ID you can get from the Organizations section of the AWS console
```
bundyfx: Thanks for the PR! This section could also detail targeting an OU with tis path (ie. /marketing/production) which does the same thing as using the OU id. We should attempt to merge all information about targeting OU's for deployments into one documentation section :)

cristim: Thanks, I was not aware of this alternative way, I'll edit the PR shortly to mention it as well. Or maybe even instead of the ID, what do you think?

sbkok: Hi @cristim, thanks for your contribution. Yes, that would have my preference too, referencing the OU path. OU-ids are hardly readable.
Currently, your changes are applied in the `admin-guide.md`, but I think the `user-guide.md` would make most sense to add these. I had to get used to this difference in the beginning too.
Maybe it would be good to highlight the difference in the docs at the top, pointing to the other doc.

sbkok:
The proposed changes in this PR are covered in the [user guide here](https://github.com/awslabs/aws-deployment-framework/blob/master/docs/user-guide.md#deployment-map).

I still think it would make sense to add references to both documents, to point users to the right documentation as discussed above. I can imagine more people run into this issue.

Considering the lack of activity, I'm closing this PR. Feel free to create a new PR or reference/reopen this one if you get time to work on it later.

Issue #202, opened 2019-12-17
Pipeline with S3 as Deployment Provider fails with Insufficient permissions


at7925:
Pipeline with S3 as Deployment Provider fails with error:
  Insufficient permissions
  You do not have sufficient permissions to call s3.putObject for the deployment bucket

Example:
```
pipelines:
  - name: poc-s3-deploy
    default_providers:
      source:
        provider: codecommit
        properties:
          account_id: 111111111111
      deploy:
        provider: s3
    targets:
      - target: 999999999999
        properties:
          bucket_name: s3-target-us-east-1-999999999999
          object_key: www
          extract: True
```
**Verifying role:**
```
aws codepipeline get-pipeline --name adf-pipeline-poc-s3-deploy | grep adf-cloudformation-role
    "roleArn": "arn:aws:iam::111111111111:role/adf-cloudformation-role"
```
**Expected**
```
"roleArn": "arn:aws:iam::999999999999:role/adf-cloudformation-role"
```
**Actual Result**
```
"roleArn": "arn:aws:iam::111111111111:role/adf-cloudformation-role"
```
**Conclusion:** incorrect account ID is used for RoleArn

PR #203, opened 2019-12-17
use target account for roleArn instead of code-source account

at7925:
*Issue #, if available:*
#202 Pipeline with S3 as Deployment Provider fails with error
*Description of changes:*
use target account for roleArn instead of code-source account for S3 deployment provider 


bundyfx:
Thanks! will go into 3.0 release :)

PR #224, opened 2020-01-27
Add S3-URI as upload path style

### LINKED ISSUE #197: Set return format for upload: function
When using the upload function in ADF the returned URL cannot be used in (for example) a CodeBuild Project as the URL format for CodeBuild requires either s3:// notation, or at least the bucket/path notation.

It would be great if the upload function could be configured to return a specific format.

Let me know if you would like to see a code sample to clarify where this could be used.
Alternatively @thomasmcgannon can provide more information as well.

sbkok:
In order to reference files from CloudFormation, like a Swagger
OpenAPI definition, you need to use the `s3://` format.

At the moment, the `upload:` helper is able to return `path` or
`virtual-hosted` urls. With this PR, it will get support to return
S3-URIs as well.

Modifications:

- Added S3-URI syntax support.
- Removed redundant code for `path` and `virtual-hosted`.
- Added tests to verify the styles and different region configs.
- Documentation has been updated to clearly distinguish the different
  output styles.

**References:**

- Fixes #197 (requests `s3://` as an output format for `upload:` helper).

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -393,6 +393,8 @@ pipelines:
 
 Parameter injection solves problems that occur with Cross Account parameter access. This concept allows the resolution of values directly from SSM Parameter Store within the Deployment account into Parameter files *(eg global.json, account-name.json)* and also importing of exported values from CloudFormation stacks across accounts and regions.
 
+#### Retrieving parameter values
+
 If you wish to resolve values from Parameter Store on the Deployment Account directly into your parameter files you can do the the following:
 
 ```yaml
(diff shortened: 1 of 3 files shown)


sbkok:
@bundyfx While writing the tests, I noticed that `virtual-hosted` style is using `http://` as its output protocol, not `httpS://`.

Is this for any particular reason that I am not aware of, or should I change that to become `https://` as well?

bundyfx:
> @bundyfx While writing the tests, I noticed that `virtual-hosted` style is using `http://` as its output protocol, not `httpS://`.
> 
> Is this for any particular reason that I am not aware of, or should I change that to become `https://` as well?

@sbkok Should be using https:// for sure! Good spotting, if you want to make that change this looks good to go, awesome work and thank you very much!

sbkok:
@bundyfx Thanks for your quick response. I've updated the PR.

Today, @thomasmcgannon shared with me that CodeBuild project source locations refer to S3 objects without the `s3://` protocol. So  I've split the setup into two styles:

- `s3-uri` which returns: `${bucket}/${key}`.
- `s3-url` which returns: `s3://${bucket}/${key}`.

PR #225, opened 2020-01-27
Version lock missing CDK dependencies

### LINKED ISSUE #223: CodeBuild failure: Newer version of CDK is required
While running ADF v2.1.2, the `aws-deployment-framework-pipelines` pipeline fails in the CodeBuild step. Similar to  #182.

It fails with a message stating:
```
A newer version of the CDK CLI (>= 1.21.0) is necessary to interact with this app
```

When it tries to execute:
```
Command did not exit successfully cdk synth --app adf-build/cdk/generate_pipeline_stacks.py 1> /dev/null exit status 1
```

---

**Root cause:**
I have found the root cause of the problem. In the list of required python modules required by: `adf-build/shared/requirements.txt`, it is missing two requirements:
```
aws_cdk.cx_api==1.16.3
aws-cdk-region-info==1.16.3
```

---

**Workaround:**
For those who are unable to upgrade to the latest release of ADF yet, apply the following patch file on your `aws-deployment-framework-bootstrap` repo:
```patch
--- adf-build/shared/requirements.txt
+++ adf-build/shared/requirements_patched.txt
@@ -49,4 +49,6 @@ aws_cdk.aws_cloudwatch==1.16.3
 aws_cdk.aws_kinesis==1.16.3
 aws_cdk.aws_s3_assets==1.16.3
 aws_cdk.aws_sns_subscriptions==1.16.3
+aws_cdk.cx_api==1.16.3
+aws-cdk-region-info==1.16.3
 aws-cdk.core==1.16.3
```

Save this patch file as `version_lock.patch` in the root of the repo.
Then execute:
```bash
patch adf-build/shared/requirements.txt < version_lock.patch && rm version_lock.patch
git add adf-build/shared/requirements.txt
git commit -m "Version lock missing CDK dependencies"
```

---

**How about the latest ADF version 3.0.3?**

This issue is not affecting anyone yet, as ADF includes a recent release of CDK in its requirements. However, it has the same problem where some dependencies of CDK are not version locked in the `requirements.txt`. So in case an organization would run ADF v3.0.3 and CDK gets updates that break the version compatibility, it will affect them too.

I will push a PR to get this fixed in the next release.

sbkok:
In the current release of ADF, the `shared/requirements.txt`
automatically includes three CDK modules with the latest version.

This commit will version-lock those modules to version `1.20` like all
other modules, such that it will not run into compatibility issues when
new versions of CDK are released.

The missing dependencies were:

- `aws-cdk.aws-sam`
- `aws-cdk.cx-api`
- `aws-cdk.region-info`

**References:**

- Fixes #223 (CodeBuild failure: Newer version of CDK is required issue)

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/requirements.txt
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/requirements.txt
@@ -50,4 +50,7 @@ aws_cdk.aws_cloudwatch==1.20
 aws_cdk.aws_kinesis==1.20
 aws_cdk.aws_s3_assets==1.20
 aws_cdk.aws_sns_subscriptions==1.20
+aws-cdk.aws-sam==1.20
+aws-cdk.cx-api==1.20
+aws-cdk.region-info==1.20
 aws-cdk.core==1.20


bundyfx:
Fantastic! Thanks!

PR #226, opened 2020-01-27
Linting on global.yml of deployment

deltagarrett:
Fixed statement for adf-codebuild-policy in the deployment global yml definition. Double PrincipalOrgID condition was failing while linting.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
@@ -229,9 +229,6 @@ Resources:
             Resource: 
               - "*"
           - Effect: Allow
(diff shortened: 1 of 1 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
```diff
@@ -229,9 +229,6 @@ Resources:
             Resource: 
               - "*"
           - Effect: Allow
-            Condition:
```
sbkok: I think these got moved in the process of merging changes.

The condition should have been applied on the "sts:AssumeRole" action listed above it.
Could you update the PR to move these lines to be part of that statement?

bundyfx:
This was part of 3.0.4.. closing.

Issue #232, opened 2020-01-29
Add support for enabling Enterprise Support for ADF Provisioned Accounts


StewartW:
**Use Case**
When we provision new accounts at the moment - We use terraform to create the account and carry out a series of tasks. One of these is enabling Enterprise Support for the newly created account. 

I think exposing a config option on the schema for accounts created via adf-accounts/*.yml would be a great feature for others and allow us to move closer to deprecating TF fully in favour of ADF.

javydekoning:
In this case, I think we should support all support levels. 

An alternative approach is running a Lambda in the master account, triggered on organization join/invite CloudTrail events. 

Or use tags on accounts to determine the support level. Together with a Lambda that runs on a schedule.

StewartW:
Hey @javydekoning  Thanks for the comments :) 
As at the moment - Only account creation is supported via ADF - I'm not sure if there's a situation in which we would be removing Support from individual accounts

My understanding of the Support tiers (obviously, you're in a better position to understand them, so correct me if I'm wrong)  is that it will work for both ES and Business tiers. I'm not sure how many people would use ADF with Developer Plans.

Likewise, I'm not sure (again, I'll defer to experience) how many customers downgrade their plans. 

We have alternative approaches already, but I feel like this would be a good problem to solve for other / potential users as well?

javydekoning:
It would be a great addition. However, if we build it 'as code' I'd like to see it serve both enabling and disabling. I would also prefer:

```
support_plan: 'none|developer|business|enterprise'
```

Over: 

```
enable_enterprise_support: True|False
```

Hence the Lambda might need some alterations. Hopefully in the future we can do this through API instead of via support case.

StewartW:
> It would be a great addition. However, if we build it 'as code' I'd like to see it serve both enabling and disabling. I would also prefer:
> 
> ```
> support_plan: 'none|developer|business|enterprise'
> ```
> 
> Over:
> 
> ```
> enable_enterprise_support: True|False
> ```
> 
> Hence the Lambda might need some alterations. Hopefully in the future we can do this through API instead of via support case.

Yes, actually, I think you could be on to something there. 

Totally agree, that the first example make's alot more sense. Happy to make those modifications. 
I think that for now It might be as simple as doing a NOOP for the ``` none|developer```  options for now (Perhaps some console output) explaining why it has no functionality and wrapping the current support request behind the ```business|enterprise``` options. 

[...]

PR #233, opened 2020-01-29
Supporting AWS Support Subscriptions

StewartW:
*Issue #, if available:* 
resolves issue #232 

*Description of changes:*
**Added** 
+ Support Class for carrying out AWS Support API Actions 
+ New property `enable_enterprise_support` for Account schema (Defaults to False, if not given) 

**Modified**
+ provisioner/main.py to call the support_session when an account is created. 

*comments*: 
Unsure about the location of the support class. I was thinking it should probably be the shared/ directory, but it imports the Account class from the provisioner source code, so I've placed it there. Let me know what you think about this. 


javydekoning:
What will happen when the bool is flipper from `True` to `False`?

StewartW:
> What will happen when the bool is flipped from `True` to `False`?

Hey, Yeah so the account would only be subscribed on creation. Updating the True|False value wouldn't have any impact. I guess this is keeping with the functionality already set by the delete_default_vpc functionality / tags i.e no revert function.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(self, role):
+        self.client = role.client("support", region_name='us-east-1', config=Support._config)
+
+    def check_aws_support_subscription(self):
+        try:
+            self.client.describe_cases()
+            return True
+        except (ClientError, BotoCoreError) as e:
+            if e.response["Error"]["Code"] == "SubscriptionRequiredException":
+                LOGGER.warning("Failed subscription for account")
```
sbkok: This warning can be removed, the info message is clear enough.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+                LOGGER.info('Enterprise Support is not enabled')
+                return False
+            raise e
```
sbkok: Could you change this to become: `raise`?

There is a big difference between `raise e` and `raise`, where `raise e` generated a new stack trace instead of reusing the previous one. See [this StackOverflow answer](https://stackoverflow.com/a/45986945).

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def enable_enterprise_support_for_account(self, account: Account, account_id):
+        try:
+            cc_email = account.email
```
sbkok: Could you add a debug log message that states that we are about to create a ticket with its context?
That makes debugging easier in the future.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+                subject=subject,
+                serviceCode='account-management',
+                severityCode='low',
+                categoryCode='billing',
+                communicationBody=body,
+                ccEmailAddresses=[
+                    cc_email,
+                ],
+                language='en',
+            )
```
sbkok: An info log message that states that a support case has been created to get enterprise support enabled on the account would be very useful. Please include the account details and add the `caseId` that is returned by the [create_case API](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/support.html#Support.Client.create_case)

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+        except (ClientError, BotoCoreError) as e:
+            LOGGER.error('Failed to enable enterprise support for account: '
+                         f'{account.full_name} ({account.alias}): {account_id}')
+            LOGGER.error(e)
```
sbkok: Could you please remove the last error and add `exc_info=True` as an argument in the first error log call.

This will enable the Python logger to capture the stacktrace.
You can add it by setting exc_info=True, like this:

```python
LOGGER.error(
    'Failed to enable enterprise support for account: '
    f'{account.full_name} ({account.alias}): {account_id}',
    exc_info=True,
)
```

Additionally, I think we might want to `raise` the error just like you did in the `check_aws_support_subscription` function such that it will not go unnoticed.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+class Support: # pylint: disable=R0904
+    """Class used for accessing AWS Support API
+    """
+
+    _config = Config(retries=dict(max_attempts=30))
+
+    def __init__(self, role):
+        self.client = role.client("support", region_name='us-east-1', config=Support._config)
+
+    def check_aws_support_subscription(self):
```
sbkok: Please add comments describing what the method does and what it returns, [like this for example](https://github.com/awslabs/aws-deployment-framework/blob/90f342c1301005596edf6b1145ecbcee1a928d4a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/retrieve_organization_accounts.py#L164-L176).

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/src/support.py
```diff
@@ -0,0 +1,58 @@
...
+            self.client.describe_cases()
+            return True
+        except (ClientError, BotoCoreError) as e:
+            if e.response["Error"]["Code"] == "SubscriptionRequiredException":
+                LOGGER.warning("Failed subscription for account")
+                LOGGER.info('Enterprise Support is not enabled')
+                return False
+            raise e
+
+    def enable_enterprise_support_for_account(self, account: Account, account_id):
```
sbkok: Please add comments to describe the method, its inputs and what possible errors it might return.

sbkok:
@javydekoning Good point!

I think we should update this in the documentation. So it clearly describes the to-be expected behavior for all of those attributes (only applies the value set at creation, not when changed after).

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/provisioner/main.py
```diff
@@ -34,25 +35,29 @@ def main():
...
+        # This only runs on account creation at the moment.
+        support_session.set_support_level_for_account(account, account_id)
+
     sts = STS()
     role = sts.assume_cross_account_role(
         'arn:aws:iam::{0}:role/{1}'.format(
             account_id,
             adf_role_name
-        ), 'delete_default_vpc'
+        ), 'adf_account_provisioning'
```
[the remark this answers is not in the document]
StewartW: Hey @sbkok. 
If you're referring to 'adf_account_provisioning' this isn't the name of the role. It will still use the adf-role (Be that the default organization access role / a custom one). 

This is only changing the role session name, to make it more generic, as this session is no longer used exclusively to delete the default vpc, and might even be used for more going forward. 

Hope this is okay, but otherwise I'm happy to revert this one line.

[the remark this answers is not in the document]
sbkok: Ah, my bad, I should have looked at those brackets more closely haha.
Thanks for your quick response and explanation, looks good.

StewartW:
> Your changes look great, thanks for developing this feature!
> One last thing: it seems that the [double newline](https://github.com/awslabs/aws-deployment-framework/pull/233/files#r406213426) is still part of the last commit you pushed.
> Could you remove that one so we can get it merged into ADF?

Apologies for the delay in fixing that. Amended the commit now.

StewartW:
Hey @sbkok  / @deltagarrett  any ETA on if we can get this merged?

PR #248, opened 2020-03-04
Fix import references of export to output key

sbkok:
In the code and docs of the import statement, it was referring to the
`export_key`. However, in the fine print it stated that it is referring
to the `output_key` as part of the Stack outputs instead.

This was confusing people to create outputs with export ids, thinking
they had to. This update renames the usage to `output_key` instead such
that it is consistent in the docs and code.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/resolver.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/resolver.py
@@ -34,16 +34,16 @@ class Resolver:
 
     def fetch_stack_output(self, value, key, optional=False): # pylint: disable=too-many-statements
         try:
-            [_, account_id, region, stack_name, export] = str(value).split(':')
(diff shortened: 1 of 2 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/resolver.py
```diff
@@ -73,7 +73,16 @@ def fetch_stack_output(self, value, key, optional=False): # pylint: disable=too-
                 self.stage_parameters[parent_key][key] = stack_output
             else:
                 if not stack_output:
-                    raise Exception("No Stack Output found on %s in %s with stack name %s and output key %s" % account_id, region, stack_name, export)
```
sbkok: The formatting of the string got interpreted wrong, resulting in the following error:
```
raise Exception("No Stack Output found on %s in %s with stack name %s and output key %s" % account_id, region, stack_name, export)
TypeError: not enough arguments for format string
```

With this PR this is fixed as well.

PR #249, opened 2020-03-04
Adds s3-key-only style

EdoBarroso:
*Issue #, if available:*
No issue available

*Description of changes:*
Added a new Upload style called `aws-key-only`. This is very similar to `s3-uri`, but it only returns the Key value.

It's very useful in case you need to use CDK to declare a Lambda Function and want to use the `fromBucket` method for the source code. 

Since you can't use CDK `fromAsset` method with ADF, then you would have to find another way to upload your Lambda code using a native method like ADF Upload. `fromBucket` method  asks for Bucket Name and S3 Key to be provided, but you can't know those before running the ADF Build stage, so you need to find a way to declare this values using CloudFormation Parameters. Since `s3-uri` returns a String similar to `<bucketName>/adf-upload/<assetFolder>/<asset>` and that's kinda difficult to separate using native CloudFormation functions, an easier approach is to use two CloudFormation parameters:

```
Parameters:
  BucketName: 'resolve:/cross_region/s3_regional_bucket/<region>'
  BucketKey: 'upload:s3-key-only:lambda/code.zip'
```

Then you could use both to put into CDK's `lambda.Code.fromBucket` code declaration

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -445,7 +445,7 @@ upload:${style}:${local_path}
 upload:${style}:${region}:${local_path}
 ```
 
-There are three different styles that one could choose from.
+There are five different styles that one could choose from.
 
 * `path` style, as shown in the example above, will return the S3 path to the object as.
   This is referred to as the classic [Path Style method](https://docs.aws.amazon.com/AmazonS3/latest/dev/VirtualHosting.html).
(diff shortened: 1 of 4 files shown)


sbkok:
I think your use case would benefit a lot of others.
It would be great to get this way of deploying lambda code using CDK documented. Could you create a separate PR with documentation and a sample how you could use this @EdoBarroso?

EdoBarroso:
> I think your use case would benefit a lot of others.
> It would be great to get this way of deploying lambda code using CDK documented. Could you create a separate PR with documentation and a sample how you could use this @EdoBarroso?

Will do, for sure. Thanks!

nerdyness:
> > I think your use case would benefit a lot of others.
> > It would be great to get this way of deploying lambda code using CDK documented. Could you create a separate PR with documentation and a sample how you could use this @EdoBarroso?
> 
> Will do, for sure. Thanks!

@EdoBarroso has this happened? If so, could you please share a link to this?

PR #251, opened 2020-03-06
fixed setting account id on deployment role

triha74:
*Description of changes:* Bugfix to setting of deployment role for deploy actions.
```
 default_providers:
      source:
        provider: codecommit
        properties:
          account_id: 123123123
      deploy:
        provider: cloudformation
        properties:
          role: deploy-role-for-this-project
```

That config would result in arn:aws:iam::None:role/deploy-role-for-this-project being set.

This fixes it so that... arn:aws:iam::123123123:role/deploy-role-for-this-project is set


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codepipeline.py
```diff
@@ -49,7 +49,7 @@ def _generate_role_arn(self):
                 return 'arn:aws:iam::{0}:role/{1}'.format(self.account_id, self.map_params["default_providers"]["build"].get('properties', {}).get("role"))
         if self.target.get('properties', {}).get('role') or self.map_params["default_providers"]["deploy"].get('properties', {}).get("role"):
             if self.category == 'Deploy':
-                return 'arn:aws:iam::{0}:role/{1}'.format(self.account_id, self.map_params["default_providers"]["deploy"].get('properties', {}).get("role"))
+                return 'arn:aws:iam::{0}:role/{1}'.format(self.target['id'], self.map_params["default_providers"]["deploy"].get('properties', {}).get("role"))
```
sbkok: I think the code in this function needs to be rewritten.
From the code on line 47, it looks like CodeBuild could potentially use this too when used to deploy.

With CodeBuild, it would need to run from inside the deployment account, in that case `self.account_id` makes sense. But as you rightly pointed out, with CloudFormation you would need the target account id instead.

So, I think it should be:
```python
account_id = self.account_id if self.provider == 'CodeBuild' else self.target['id']
specific_role = self.target.get('properties', {}).get('role') or self.map_params['default_providers'][self.category.lower()].get('properties', {}).get('role')
if specific_role:
  return 'arn:aws:iam::{0}:role/{1}'.format(account_id, specific_role)
```

triha74: I did that.. but I think we actually need more refactoring... there are sooo many `self.map_params.get('default_providers', {}).get(provider_name, {}).get('properties', {}).get(property)` would be nice to have the providers and targets as objects with getters

bundyfx: Indeed those changes are coming shortly but for now lets just get this working as intended.

EdoBarroso:
@sbkok can you merge it?

Issue #272, opened 2020-06-08
Issue with paginator not being used in moved_to_root and direct access to key


rickardl:
Hi,

When moving one of our larger accounts back to the root OU we get the following error in one of the step functions (that invoked moved_to_root). Our SSM Parameters has empty Descriptions and it has over 100 entries and this becomes an issue with how the current implementation handles some edge cases.

```
[ERROR] KeyError: 'Description'
Traceback (most recent call last):
  File "/var/task/moved_to_root.py", line 93, in lambda_handler
    execute_move_action(action, account_id, parameter_store, event)
  File "/var/task/moved_to_root.py", line 76, in execute_move_action
    return remove_base(account_id, regions, role, event)
  File "/var/task/moved_to_root.py", line 65, in remove_base
    thread.join()
  File "/opt/python/thread.py", line 30, in join
    raise self.exc
  File "/opt/python/thread.py", line 22, in run
    **self._kwargs
  File "/var/task/moved_to_root.py", line 31, in worker_thread
    parameters = [param['Name'] for param in parameter_store.client.describe_parameters()['Parameters'] if 'Used by The AWS Deployment Framework' in param['Description']]
  File "/var/task/moved_to_root.py", line 31, in <listcomp>
    parameters = [param['Name'] for param in parameter_store.client.describe_parameters()['Parameters'] if 'Used by The AWS Deployment Framework' in param['Description']]
```

Digging into the deployment framework we can see that `moved_to_root.py` has the following issues in the `worker_thread`

* it does not paginate the API response
* it doesn't take into the consideration that the Description key for a `parameter` might be empty, therefore the list comprehension can fail with an error.

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues with the latest release.

PR #273, opened 2020-06-08
Support paginator for parameters and empty descriptions in moved to root lambda 

rickardl:
The below change hopefully resolves all issues we encountered in #272 

* support larger number of entries of SSM parameters by using the paginator method of operation (using the paginator helper)
* being able iterate over empty descriptions, by handling the Description issue correctly (and defaulting to empty string)

- [X] By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


src/lambda_codebase/moved_to_root.py
```diff
@@ -28,9 +29,10 @@ def worker_thread(sts, region, account_id, role, event):
         'remove_base')
 
     parameter_store = ParameterStore(region, role)
-    parameters = [param['Name'] for param in parameter_store.client.describe_parameters()['Parameters'] if 'Used by The AWS Deployment Framework' in param['Description']]
-    for parameter in parameters:
-        parameter_store.delete_parameter(parameter)
+    for page in paginator(parameter_store.client.describe_parameters):
```
sbkok, line 32: Could you please change this to use the Boto3 paginator instead?
See the docs in boto3 here: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html

sbkok (changes requested):

src/lambda_codebase/moved_to_root.py
rickardl: Absolutely, I'll update the PR next week!

sbkok (approved):
Thanks for updating the pull request and contributing this fix.
Looks good to me.

deltagarrett (approved):
LGTM

PR #275, opened 2020-06-18
Fix supported list of intrinsic upload path styles

sbkok:
**Why?**

When the `Resolver` would find a usage of an intrinsic upload, it would check
the requested path style against an outdated list of supported path
styles.

**How did this happen?**

The `S3` class implements the path styles, but did not maintain the list
of supported paths. While new paths were added to the S3 class, its
supported list of paths in the `Resolver` class was not updated.

With this change, the supported list of path styles becomes a class
method on the `S3` class instead. Such that the list and its
implementation are managed by the same class instead.


thomasmcgannon (approved):
lgtm

kalleeh (approved):
lgtm

Issue #282, opened 2020-07-16
Should ADF implement CDK Pipelines?


EdoBarroso:
There’s a new CDK feature called CDK Pipelines: https://aws.amazon.com/blogs/developer/cdk-pipelines-continuous-delivery-for-aws-cdk-applications/

Should ADF adopt that as the way of dealing with new pipelines, using the deployment-map declarations?

thomasmcgannon:
This would be a valuable addition, have you had any thoughts on how something like this could be integrated?
Would you be open to contribute on work like this?

sbkok:
I'm closing this issue as it has been inactive for a long time.

Please reopen if you are interested in contributing this or if there is a general interest into having this part of ADF.

Thank you!

Issue #185, opened 2019-11-15
Reintroduce a pipeline type 


StewartW:
Hey, 
We've got a business requirement for having some pipelines that have drastically different shapes. 
As such we've made some changes to adf-build/shared/cdk/cdk_stacks/main.py (and schema) 

[example_snippet.txt](https://github.com/awslabs/aws-deployment-framework/files/3850961/example_snippet.txt)

In the example snippet, you can see that we have two skyscanner specific pipelines, these are currently situated in adf/shared/cdk/cdk_stacks/skyscanner_stacks.py.
(Ideally, the dream would be to install these as part of the requirements.txt)
I think a touch would be if, what we call, the adf_default pipeline was also situated outside of main.py and imported in. 

I think it would be quite beneficial for other users as it means that any updates to the default pipeline won't require extensive merge conflicts, and gives it a certain flexibility that would probably be beneficial to most ?

Let me know what you think?  This is something I'm happy to work together on if you think it needs refined. 

Edit: I've got a branch with how I think this would look.

bundyfx:
Thanks for opening the issue, I think there is something we can do further to make this more modularized for sure. Im going to have a think/discussion about this one a bit more and take your example into mind. Will update this shortly with some more thoughts but I think you are on the right train of thought here.

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues with the latest release.

PR #285, opened 2020-08-02
Adding in a pipeline type param.

StewartW:
*Issue #, if available:* #185 

*Description of changes:*
I've re-introduced a pipeline type param, I hope you find this useful. The primary motivation is to make the overall shape of pipelines more configurable. For example we currently have pipelines at Skyscanner that are used for deploying CloudCustodian to all of our accounts, or executing CDK in them. (And I would like to eventually share some of these back to the ADF community at some point too.) 

I also think it is a good starting place to resolving issue #282 by creating a CDK pipeline type, and utilising the new CDK Pipeline Construct. 


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_stacks/adf_standard_pipelines.py
```diff
@@ -0,0 +1,260 @@
...
+
+    if stack_input["input"].get("params", {}).get("notification_endpoint"):
+        stack_input["input"]["topic_arn"] = adf_notifications.Notifications(
+            scope, "adf_notifications", stack_input["input"]
+        ).topic_arn
+
+    _source_name = generate_source_stage_for_pipeline(_stages, scope, stack_input)
+    generate_build_stage_for_pipeline(_stages, scope, stack_input)
+    print(stack_input["input"].get("environments", {}).get("targets", []))
+    for index, targets in enumerate(stack_input["input"].get("environments", {}).get("targets", [])):
```
sbkok, line 39: Would it be easy to move the target generation to its own function too?
In that case this function will become nothing else than setting up the notification endpoint, rendering source, build, and targets. Plus generating the related Pipeline at the end + webhooks for Github.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_stacks/adf_standard_pipelines.py
```diff
@@ -0,0 +1,260 @@
...
+                stack_input["ssm_params"][ADF_DEPLOYMENT_REGION]["kms"],
+                stack_input["input"],
+                {},  # Empty target since this is a build only stage
+            ).build
+        )
+    elif "jenkins" in _build_name:
+        _stages.append(adf_jenkins.Jenkins(scope, "build", stack_input["input"]).build)
+
+
+def generate_source_stage_for_pipeline(_stages, scope, stack_input):
```
sbkok, line 248: Could you move the functions in the order as they will be executed, so source first, build second, targets third, actions fourth, etc?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_stacks/adf_standard_pipelines.py
```diff
@@ -0,0 +1,260 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sbkok, line 1: On the filename, at the moment, it is called `adf_standard_pipelines.py` whereas you called the Pipelines Type "Default", maybe good to relate the two. Plus the name is plural: `pipelines`, whereas it is used to generate a single pipeline. What do you think about calling it `adf_default_pipeline.py` instead?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_stacks/tests/test_pipeline_creation.py
```diff
@@ -0,0 +1,142 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sbkok, line 1: Thanks for adding tests for some of the changes you made, I appreciate this!

sbkok (commented):
Thanks @StewartW for adding these changes and refactoring as you go.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_stacks/adf_standard_pipelines.py
StewartW: Makes sense. 👍

StewartW:
@sbkok  thanks for pulling in the latest patches. One less thing of my to-do list.

sbkok (approved):
Looks great @StewartW, thanks for adding this.

thomasmcgannon (approved):
LGTM

PR #288, opened 2020-09-22
Seperated out pipeline cleanup from input generation

StewartW:
*Issue #, if available:*

*Description of changes:*
Added a new python script for removing stale entries from SSM and
deleting cloudformation templates for pipelines no longer defined
in the deployment map(s)

This script is called in the post_build stage of the pipeline
generation CodeBuild project.


StewartW:
Suggesting this because it doesn't seem right to me that a script name "generate_pipeline_input" is also responsible for deleting pipelines as well. Hopefully this makes sense.

sbkok (approved):
Good suggestion, thanks for contributing these changes.

thomasmcgannon (approved):
lgtm

PR #293, opened 2020-09-29
Fix CodeBuild provider spec_inline support

sbkok:
**What was the problem?**

1. The `spec_inline` property was documented and type checked as a string.
   Whereas the CDK function used to include the `spec_inline` expects an
   object instead.

2. If CodeBuild was used as a deployment provider, it would not be able
   to use the `spec_inline`. Nor would you be able to specify
   `spec_inline` as part of a target definition. Even though the schema
   validation made it look like it would support `spec_inline`.

**What has been changed?**

* Updates the schema validation to support objects, as required
  by the CDK BuildSpec construct.
* Changes the CodeBuild provider for deployment and build stages, such that
  both support inline and file reference usage of BuildSpec definitions.
* When both the `spec_inline` and `spec_filename` properties are defined as
  part of the default provider or target properties, an error message will be
  thrown.
* Tests have been added to verify that the added method works as
  intended.


thomasmcgannon (approved):
lgtm

deltagarrett (approved):
LGTM

PR #295, opened 2020-09-30
Fix specific role usage to be used in Build and Deploy only

sbkok:
**Why?**
A bug prevented stages like Approvals to be included in a pipeline. As
the _generate_role_arn function tried to retrieve the specific role for
the approval stage, while that stage doesn't support such properties.

**What changed?**
Added a check to look for specific roles in Build and Deploy stages
only. All other stages should not use a specific role.


deltagarrett (approved):
LGTM

kalleeh (approved):
lgtm

Issue #324, opened 2021-02-02
Deprecating botocore.vendored.requests


hynynen:
Hello!

We have been receiving this warnings. How is this affected to ADF? We had advise:
Lambda function must be "refreshed" so that it pulls the latest version of cfnresponse.
"If you are using the cfn-response module for a Lambda-backed custom resource, you must modify your Lambda function resource’s Code or Runtime property and update your stack(s) in order to get the latest version of the cfn-response module which removed its dependency on botocore.requests."

Do we need to manually update ADF Lambdas somehow? Any new releases coming out before April 1st?

engrun:
Good catch @hynynen .  Hopefully ADF is not affected

sbkok:
Hi @hynynen,

Thanks for raising this as an issue.
I've checked the code that we depend on and found one reference.
Luckily this was in the samples folder, the sample-expunge-vpc one to be precise.

I have created a fix for this and will open that as a separate PR.

Best regards, Simon

PR #326, opened 2021-02-05
Remove dependency on botocore.vendored.requests

sbkok:
### Why?

Releases of botocore will no longer support the requests library as part
of its vendored libraries. The samples/sample-expunge-vpc code was
relying on this dependency indirectly.

### What?

The samples/sample-expunge-vpc code included the crhelper library
directly. This change will remove the dependency from being committed
into this repository and install it on the fly instead.

The latest update of the crhelper is removing the dependency of the
botocore.vendored.requests library. So by changing this to
update/install on the fly it will resolve issue #324.

---


thomasmcgannon (approved):
lgtm

deltagarrett (approved):
LGTM

Issue #318, opened 2021-01-21
S3 ETL fails when "Build: enabled: False


psgivens:
An example such as the following fails because the source output artifact doesn't match the deploy input artifact. I've provided the solution in this ticket. 

    - name: sample-s3-pipeline
      default_providers:
        source:
          provider: s3
          properties:
            bucket_name: packer-bucket-test
            object_key: input.zip
            account_id: 44444444444444
        build:
          enabled: False # if you wish to disable the build phase in a pipeline
        deploy:
          provider: s3
      targets:
        - target: 222222222222
          properties:
            bucket_name: test-adf-bucket-eu
            object_key: output.zip

[S3 output artifact is "output-source".](
https://github.com/awslabs/aws-deployment-framework/blob/31f4719f83e80bd165b1567fed27de8eff090460/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codepipeline.py#L324)

        if self.category == 'Source':
            action_props["output_artifacts"] = [
                _codepipeline.CfnPipeline.OutputArtifactProperty(
                    name="output-source"
                )
            ]

[When Build is set to "enabled: false", CodeBuild is skipped and the S3 deploy's input artifact is "{0}-build".format(self.map_params['name']](https://github.com/awslabs/aws-deployment-framework/blob/31f4719f83e80bd165b1567fed27de8eff090460/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codepipeline.py#L305)

        if self.category == 'Deploy':
            action_props["input_artifacts"] = [
                _codepipeline.CfnPipeline.InputArtifactProperty(
                    name="{0}-build".format(self.map_params['name'])
                )
            ]

The generated CloudFormation for the pipeline fails with a complaint that there isn't an output format for "{0}-build". 

**Solution**: Add a build the following block when [building the Deploy action after the initial set block.](https://github.com/awslabs/aws-deployment-framework/blob/31f4719f83e80bd165b1567fed27de8eff090460/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codepipeline.py#L311). 

            if not self.map_params.get('default_providers', {}).get('build', {}).get('enabled', True):
                action_props["input_artifacts"] = [
                    _codepipeline.CfnPipeline.InputArtifactProperty(
                        name="output-source"
                    )
                ]

[...]

Issue #236, opened 2020-01-31
Build stage configuration flag "enabled" for provider codebuild not working properly 2


lordoflairs:
Using pipeline with just source codecommit and then deploy with cloudformation by setting enabled to false for the build step will generate output from source not match input to cloudformation. (InputArtifacts to cloudformation does not match OutputArtifacts from previous step.)

The problem is probably in adf_codepipeline.py but I am not bright enough to figure it out. (Not a Python coder...)

E.g. changeset will be:

```
        "Stages": [
          {
            "Actions": [
              {
                "ActionTypeId": {
                  "Category": "Source",
                  "Owner": "AWS",
                  "Provider": "CodeCommit",
                  "Version": "1"
                },
                "Configuration": {
                  "BranchName": "master",
                  "RepositoryName": "myRepo",
                  "PollForSourceChanges": false
                },
                "Name": "source",
                "OutputArtifacts": [
                  {
                    "Name": "output-source"
                  }
                ],
                "Region": "eu-west-1",
                "RoleArn": "arn:aws:iam::XXXXXX:role/adf-codecommit-role",
                "RunOrder": 1
              }
            ],
            "Name": "Source-XXXXXX"
          },
          {
            "Actions": [
              {
                "ActionTypeId": {
                  "Category": "Deploy",
                  "Owner": "AWS",
                  "Provider": "CloudFormation",
                  "Version": "1"
                },
                "Configuration": {
                  "ActionMode": "CHANGE_SET_REPLACE",
                  "StackName": "adf-stack-name",
                  "ChangeSetName": "adf-stack-name",
                  "TemplateConfiguration": "stack-name-build::params/CloudCI_eu-west-1.json",
                  "Capabilities": "CAPABILITY_NAMED_IAM,CAPABILITY_AUTO_EXPAND",
                  "RoleArn": "arn:aws:iam::XXXXXX:role/adf-cloudformation-deployment-role",
                  "TemplatePath": "stack-name-build::template.yml"
                },
                "InputArtifacts": [
                  {
                    "Name": "stack-name-build"
                  }
                ],
                "Name": "CloudCI-eu-west-1-create",
                "Region": "eu-west-1",
                "RoleArn": "arn:aws:iam::XXXXXX:role/adf-cloudformation-role",
                "RunOrder": 1
              },
```

lordoflairs:
Maybe there are more issues to this...
[...]

PR #334, opened 2021-03-03
Fix usage of source artifacts in deploy stages when build stage is disabled

sbkok:
## Why?

When the build stage is disabled, the deploy stages incorrectly tried to use the build output artifact.
These should use the source artifact instead.

## Linked issues

* #236
* #318


thomasmcgannon (approved):
lgtm!

deltagarrett (approved):
LGTM

PR #336, opened 2021-03-11
Fix resolve intrinsic function use in first parameter

sbkok:
**Why?**

A bug prevented the use of the `resolve:` intrinsic function on the first parameter in in the first parameter file.

**What?**

The parameters were initialized as an empty dictionary. While reading a file that does not exist would return a predefined object instead. The fix uses the later instead of an empty dictionary.


deltagarrett (approved):
LGTM

thomasmcgannon (approved):
lgtm

PR #341, opened 2021-03-26
Update Jinja2 to v2.11.3

sbkok:
**Why?**

A vulnerability was reported on v2.11.2 of Jinja2, as reported per
CVE-2020-28493. Since we make use of this dependency, we need to link to
the version that resolved this issue.

Since ADF does not use the `urlize` that is impacted, it is not at risk.
Hence it is not required to release a new version or update your ADF
installation.

**What?**

Updated to 2.11.3.


deltagarrett (approved):
LGTM

thomasmcgannon (approved):
LGTM

Issue #342, opened 2021-03-26
Allow different names for the SAR application


benbridts:
[The documentation says:](https://github.com/awslabs/aws-deployment-framework/blob/master/docs/serverless-application-repo.md)
> Application Name
> The stack name of this application created via AWS CloudFormation. This should always be aws-deployment-framework.

However, if you want to install ADF using CloudFormation / SAM, there is no way to select the name . Instead the resulting (nested) CloudFormation stack will have a name like `${parentStack}-${resourceName}-${randomId}`. Eg. if you deploy the template below as 'adf-test', it might look like `adf-test-AwsDeploymentFrameWork-W4AUMAFJESS7`.

I looked at the resulting template, and I couldn't find any use of `AWS::StackName` in it. This leads me to a few questions:
- Is it really needed to use `aws-deployment-framework` as the name?
- If not, can we update the documentation?
- Is there anything that would block using CloudFormation to deploy the SAR app?

```yaml
Description: AWS Deployment Framework SAR app and related resources
Parameters:
  Version:
    Description: ADF Version
    Type: String
    Default: 3.1.2
Transform: AWS::Serverless-2016-10-31
Resources:
  AwsDeploymentFrameWork:
    Type: AWS::Serverless::Application
    Properties:
      Location:
        ApplicationId: arn:aws:serverlessrepo:us-east-1:112893979820:applications/aws-deployment-framework
        SemanticVersion: !Ref Version
```

benbridts:
Update: I found 1 hardcoded values in the template that depends on this naming (searched for `serverlessrepo-aws`):

https://github.com/awslabs/aws-deployment-framework/blob/master/src/template.yml#L414

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues with the latest release.

Issue #332, opened 2021-02-19
Basic GovCloud Support


dsudduth:
# GovCloud Support

## Summary

This feature request would provide basic support for GovCloud regions by adjusting partitions throughout templates and the code base to leverage the `${AWS::Partition}` [pseudo parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html). It would also add conditional statements for services that do not exist in GovCloud. *i.e., Tagging Policies*.

## Motivation

GovCloud customers are looking for a robust CI/CD framework based on AWS native tooling. ADF is one of the most feature rich options. With small changes this framework could easily be adapted to support a wider audience.

## In Scope

- Modify the code base to support the `arn:aws-us-gov` GovCloud partition.
- Provide support for the `us-gov-west-1` region as  a primary install location.
- Avoid failures due to missing features in GovCloud. The only item identified in the code base is Tagging Policies.

## Out of Scope

**Account Vending in GovCloud** - Account vending in GovCloud requires accounts to be created in commercial regions and linked. The changes needed to support this would be significant and will not be addressed by this feature.

## References

- [ARN Resource Names](https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-arns.html)
- [Pseudo Parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html)

sbkok:
Thanks for submitting this request @dsudduth!
I'm looking into the required steps and will reach out to you directly to sync up.

OlafConijn:
interesting! we are currently building govcloud support in org-formation (https://github.com/org-formation/org-formation-cli). if you find it interesting feel free to ping me and we can compare noted. easiest way to do so is over the org-formation slack (https://join.slack.com/t/org-formation/shared_invite/enQtOTA5NjM3Mzc4ODUwLTMxZjYxYzljZTE5YWUzODE2MTNmYjM5NTY5Nzc3MzljNjVlZGQ1ODEzZDgyMWVkMDg3Mzk1ZjQ1ZjM4MDhlOGM).

dsudduth:
Hi @OlafConijn! Thanks for sharing your tool. Looks great! I'll connect with you on Slack. Any tool that can work across commercial and GovCloud is of interest.

sbkok:
[...]

PR #343, opened 2021-03-31
Improvements to the SAR app template

benbridts:
This improves the SAR Application template, I used three commits to make review easier and allow cherry-picking if that would be preferred:

Improve SAR Application template (These are things I discovered by running `cfn-lint -e -cI`):
- replaces every `arn:aws:` with `arn:${AWS::Partition}` (should help with #332 )
- replaces `!Join ['', ...]` with `!Sub`
- added `UpdateReplacePolicy` if there was a `DeletionPolicy`
- changed the type of properties where it didn't match the specification
- removed an unused `!Sub`

Remove invalid permissions: `s3:Get` (without a wildcard) doesn't give any permissions

Use references where possible:
There were still some hardcoded values in the template. I removed those where it was a relatively small change. This should make #342 possible, and make it safer to change names of resources the future.
There are some hardcoded values I didn't change:
- If the resources are created outside of this template
-  `project/aws-deployment-framework-base-templates`, because removing that circle would require converting a managed policy to an inline policy, which is a bigger change


src/template.yml
```diff
@@ -68,21 +68,22 @@ Resources:
               StringEquals:
                 aws:PrincipalOrgID: !GetAtt Organization.OrganizationId
             Resource:
-              - !Sub arn:aws:s3:::${BootstrapTemplatesBucket}
-              - !Sub arn:aws:s3:::${BootstrapTemplatesBucket}/*
+              - !Ref BootstrapTemplatesBucket
```
sbkok, line 71: According to the [S3 Policy docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html), the resource should refer to the Arn of a bucket in S3 Bucket Policies. When you use Ref, according to the [S3 CFn docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html#aws-properties-s3-bucket-return-values), it returns the bucket name instead.

I think this line should become:
```yaml
!GetAtt BootstrapTemplatesBucket.Arn
# Or:
!Sub "${BootstrapTemplatesBucket.Arn}"
```

src/template.yml
```diff
@@ -68,21 +68,22 @@ Resources:
...
+              - !Sub "${BootstrapTemplatesBucket.Arn}/*"
             Principal:
               AWS: "*"
           - Action:
               - s3:PutObject*
             Effect: Allow
             Resource:
-              - !Sub arn:aws:s3:::${BootstrapTemplatesBucket}
-              - !Sub arn:aws:s3:::${BootstrapTemplatesBucket}/*
+              - !Ref BootstrapTemplatesBucket
```
sbkok, line 79: Same here, use `!GetAtt BootstrapTemplatesBucket.Arn` instead.

src/template.yml
```diff
@@ -321,7 +319,6 @@ Resources:
               - "codecommit:List*"
               - "codecommit:CancelUploadArchive"
               - "codecommit:UploadArchive"
-              - "s3:Get"
```
sbkok, line 324: Yes, that one is not adding any value, thanks for removing.

src/template.yml
```diff
@@ -411,15 +408,17 @@ Resources:
               - "s3:ListBucket"
               - "s3:PutObject"
             Resource:
-              - "arn:aws:s3:::serverlessrepo-aws-deplo-bootstraptemplatesbucket-*"
-              - "arn:aws:s3:::serverlessrepo-aws-deplo-bootstraptemplatesbucket-*/*"
-              - "arn:aws:s3:::adf-shared-modules-*-*"
-              - "arn:aws:s3:::adf-shared-modules-*-*/*"
+              - !Ref BootstrapTemplatesBucket
```
sbkok, line 411: Same as above, using `!GetAtt BootstrapTemplatesBucket.Arn`.

src/template.yml
```diff
@@ -801,15 +795,27 @@ Resources:
...
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action: ssm:GetParameter
+            Resource:
+              - !Sub "arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${SharedModulesBucketName}"
+              - !Sub "arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/deployment_account_region"
+      Roles:
+        # Created by the Serverless Transform of CrossRegionBucketHandler
+        - !GetAtt CrossRegionBucketHandlerRole
```
sbkok, line 818: Personally I don't like relying on a to-be-generated transform resource.
This makes reviewing the code very hard to follow.
If this is the only way to break-out of the circular reference, I would rather move the other policies away from the lambda function into a separate role definition as well.

Plus I think this would need to be a `!Ref` instead of a `!GetAtt`.

sbkok (changes requested):
Thanks for cleaning up the template!
I've added a few comments that I think should be improved before we can merge this one.

src/template.yml
benbridts: I'll keep the hardcoded name for now, making it a dependency can be done in a different PR

src/template.yml
```diff
@@ -434,16 +434,18 @@ Resources:
               - "s3:ListBucket"
               - "s3:PutObject"
             Resource:
-              - !Sub "arn:${AWS::Partition}:s3:::serverlessrepo-aws-deplo-bootstraptemplatesbucket-*"
-              - !Sub "arn:${AWS::Partition}:s3:::serverlessrepo-aws-deplo-bootstraptemplatesbucket-*/*"
+              - !GetAtt "BootstrapTemplatesBucket.Arn"
+              - !Sub "${BootstrapTemplatesBucket.Arn}/*"
```
benbridts: These lines might not have made it in the first version of the PR

benbridts:
@sbkok  I think I fixed everything, I also made a quick overview of the changes that actually reduce permissions:

- L439-440: `arn:aws:s3:::serverlessrepo-aws-deplo-bootstraptemplatesbucket-*` replaced with `BootstrapTemplatesBucket.Arn `
- L443-444: `arn:aws:s3:::adf-shared-modules-*-*` replace with `arn:${AWS::Partition}:s3:::${SharedModulesBucket.BucketName}`
- L833: `arn:aws:s3:::adf-shared-modules-*` replaced with `arn:${AWS::Partition}:s3:::adf-shared-modules-${DeploymentAccountMainRegion}-*`

src/template.yml
```diff
@@ -830,12 +825,15 @@ Resources:
                 - s3:PutEncryptionConfiguration
                 - s3:PutBucketPolicy
                 - s3:PutBucketPublicAccessBlock
-              Resource: "arn:aws:s3:::adf-shared-modules-*"
+              # This must match BucketNamePrefix of the SharedModulesBucket resource
+              Resource: !Sub "arn:${AWS::Partition}:s3:::adf-shared-modules-${DeploymentAccountMainRegion}-*"
```
sbkok, line 829: When updating, I noticed that the DeploymentAccountMainRegion is not enforced.
If you forget to specify it, it will create a new bucket without the region in its name.
So far, that has not caused any issues. However, when we introduce this change, it will.

So either we should enforce the DeploymentAccountMainRegion in the stack, and make it clear that users should not forget to specify the region. Or we should implement a way to lookup the previous value.
Unfortunately the SAR does not show the values used previously when deploying the same template.

sbkok (changes requested):

sbkok (approved):
I have updated this PR to reflect the latest changes from our main branch + updated some of the issues that were flagged before but not resolved yet.

It looks good to me to merge. Thanks for contributing.

StewartW (approved):
Looks great. Really tidies up the template 👍

PR #345, opened 2021-04-01
Remove not imported CDK requirements

benbridts:
CDK documentation recommends only specifying top-level dependencies ([link](https://docs.aws.amazon.com/cdk/latest/guide/work-with-cdk-python.html#python-managemodules)).

I removed all modules that weren't imported.

The main reason for doing this, is that there is a dependency conflict. It does not throw an error if you use pip, but using pipenv, you get:

```
~$ pipenv install -r requirements.txt
[...]
Resolving dependencies...
✘ Locking Failed!
[...]
ERROR: Could not find a version that matches jsii<1.20.0,<2.0.0,>=1.15.0,>=1.16.0,>=1.26.0 (from -r [...]
There are incompatible versions in the resolved dependencies:
  jsii<1.20.0,>=1.16.0 (from -r /var/folders/pq/cvfgj3_s0c565_sc9rwywd0w0000gn/T/pipenv9fbqfx4nrequirements/pipenv-kx3l0_sv-constraints.txt (line 47))
  jsii<2.0.0,>=1.15.0 (from aws-cdk.assets==1.88.0->-r /var/folders/pq/cvfgj3_s0c565_sc9rwywd0w0000gn/T/pipenv9fbqfx4nrequirements/pipenv-kx3l0_sv-constraints.txt (line 54))
[... repeated with 1.15.0 for other cdk dependencies ...]
  jsii<2.0.0,>=1.26.0 (from constructs==3.3.71->aws-cdk.core==1.88->-r /var/folders/pq/cvfgj3_s0c565_sc9rwywd0w0000gn/T/pipenv9fbqfx4nrequirements/pipenv-kx3l0_sv-constraints.txt (line 51))
```


benbridts:
Actually there are cases where pip does throw errors, using this codebuild file:
```yaml
version: 0.2
phases:
  install:
    runtime-versions:
      python: 3.9
    commands:
      # These three lines are needed for ADF
      - aws s3 cp s3://$S3_BUCKET_NAME/adf-build/ adf-build/ --recursive --quiet
      - pip install -r adf-build/requirements.txt -q
```
I get this output

```shell
[Container] 2021/04/02 20:22:48 Running command pip install -r adf-build/requirements.txt -q
--
21 | constructs 3.3.71 has requirement jsii<2.0.0,>=1.26.0, but you'll have jsii 1.19.0 which is incompatible.
22 | awscli 1.18.140 has requirement PyYAML<5.4,>=3.10; python_version != "3.4", but you'll have pyyaml 5.4.1 which is incompatible.
23 | You are using pip version 18.1, however version 21.0.1 is available.
24 | You should consider upgrading via the 'pip install --upgrade pip' command.
25 |  
```

sbkok:
Hi @benbridts,

In the past we faced multiple issues when we only specified the direct imports.
The jsii limit was introduced by https://github.com/awslabs/aws-deployment-framework/commit/21be1c139be4c9e32bdeda570106420afe5a4975. As a jsii library version release was shown incompatible with CDK 1.88.
With the recent update to CDK v1.105 I removed that dependency the way you suggested, see #364.

In terms of CDK libraries that we depend upon indirectly, these get automatically updated to the latest version and installed if you don't specify them. Even if that means they are incompatible, as can be seen in issue #223 for example.
In a normal stack that would be okay, as you can upgrade those quickly. With ADF however, we would need to release a new version every time.

We will keep these dependency version locks, as we don't want to introduce and force every organization to upgrade ADF with the rate that CDK pushes out new releases.

I'm closing this PR as the jsii fix that was introduced is also part of #364.
Let me know if you have any thoughts on this.

Best regards,
Simon

PR #347, opened 2021-04-02
Allow top-level keys starting with x- or x_ in deployment maps

benbridts:
This allows users to define anchors in one place.

ADF already encourages anchors in the documentation, this should make them even more convenient.

I added an example in the user guide to document this feature.


sbkok (approved):
Great feature to have, thanks for contributing this!

deltagarrett (approved):
LGTM

PR #349, opened 2021-04-12
Adding support for CodeBuild to pull from docker hub

pozeus:
Adding support for CodeBuild to pull images from docker hub using LinuxBuildImage classmethod from_docker_registry (https://docs.aws.amazon.com/cdk/api/latest/python/aws_cdk.aws_codebuild/LinuxBuildImage.html#aws_cdk.aws_codebuild.LinuxBuildImage.from_docker_registry).
I added some documentation and an example as well, please feel free to change if you think it's not explanatory enough.

Images can be provided like:
```yaml
pipelines:
  - name: example-custom-image
    default_providers:
      source:
        ...
      build:
        provider: codebuild
        properties:
          image: bitnami/mongodb
    targets:
      - ...
```


sbkok:
Hi @pozeus,

Thanks for your contribution here.
I like adding support for this!

I would like to propose a slight modification so we can support other registries later as well.
At the moment, you check whether the image is defined for use with CodeBuild.
If not, it does a look-up for a docker container instead.

Could you change this to check if the string starts with: "docker-hub://${image-goes-here}"?
So we can add other registries in the future too? For example, adding support for the AWS ECR Public gallery, as is described [here](https://aws.amazon.com/about-aws/whats-new/2020/12/announcing-amazon-ecr-public-and-amazon-ecr-public-gallery/). I'm not asking to add support for the latter right away, but by prefixing the images we don't have to introduce a breaking change later on.

Let me know if it is unclear or if I can support you with these changes.

Many Thanks,
Simon

sbkok (changes requested):
Hi @pozeus,

Could you also update the PR to include the following statement?
This is specified as part of the template of our PRs and is required for all contributions to our repository.

> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Many thanks,
Simon

sbkok:
Dear @pozeus,

I put the "on hold" tag on the pull request, as we cannot merge this fix before you agree to the licensing terms.
Additionally, there are a few review comments that suggest changing the implementation slightly.
Could you look at both of these points, otherwise I will have to close the pull request without merging unfortunately.
Considering the usefulness of the feature, I think that would not be ideal.

If you don't have time to work on the fix, please agree with the licensing terms and state so here, so we can work on the improvements instead.

Best regards,
Simon

pozeus (commented):
Sorry for the delay, I will fix this coming week.

pozeus:
@sbkok Hi Simon,
Please see the latest update. If you think there is other better way's of doing this, please provide feedback.

docs/providers-guide.md
```diff
@@ -230,6 +230,15 @@ Provider type: `codebuild`.
...
   > Along with `repository_arn`, we also support a `tag` key which can be used
   > to define which image should be used (defaults to `latest`).
+  > 
+  > Image can also take an object that contains a reference to a
+  > public docker hub image with a prefix of `docker-hub://`, such as
+  > `docker-hub://bitnami/mongodb`. This allows your pipeline
+  > to consume a public docker hub image if required.
+  > Along with the docker hub image name, we also support using a tag which can
+  > be provided after the docker hub image name to define which image should
+  > be used (defaults to `latest`).
```
sbkok, line 240: Could you add an example that specifies a tag as well?

sbkok (changes requested):
Thanks for making these improvements @pozeus.

The CI build process reported the following linting issues:
```
src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py:161:0: C0303: Trailing whitespace (trailing-whitespace)

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py:165:65: C0303: Trailing whitespace (trailing-whitespace)

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py:196:58: C0303: Trailing whitespace (trailing-whitespace)

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py:163:4: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
```

Could you fix those too, so we can merge the changes after?
You can run the test suite + linting by running `tox` locally too.

pozeus:
@sbkok I've now fixed the return which was missing in the function and added the suggested example. Please review 😊

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py
```diff
@@ -159,6 +159,17 @@ def determine_build_spec(codebuild_id, default_props, target=None):
...
+    @staticmethod
+    def get_image_by_name(specific_image: str):
+        if hasattr(_codebuild.LinuxBuildImage,
+            (specific_image or DEFAULT_CODEBUILD_IMAGE).upper()):
+            return getattr(_codebuild.LinuxBuildImage,
+                (specific_image or DEFAULT_CODEBUILD_IMAGE).upper())
+        if specific_image.startswith('docker-hub://'):
+            specific_image = specific_image.split('docker-hub://')[-1]
+            return _codebuild.LinuxBuildImage.from_docker_registry(specific_image)
+        return specific_image
```
sbkok, line 171: Could you make it raise an error when we hit this line instead?

At the moment, it would return the CodeBuild image in both other return types.
While it returns a string when it is unable to find/match.

You can raise an exception that states that the CodeBuild image could not be found.
Please include the specific_image in the error message so it will be easy to debug.

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py
pozeus: Good lord I was tired yesterday 😅 
Will something like this raise exception work for you?

pozeus:
@sbkok Any comment regarding this?

sbkok (dismissed):
Perfect. Thanks for making these changes and contributing this feature!

sbkok (approved):

StewartW (approved):
Looks good to me. :shipit:

Issue #270, opened 2020-06-05
Parallel executions of aws-deployment-framework-bootstrap-pipeline leads to failure in CodeBuild


mdaehnert:
# Description
When a change is made to _CodeCommit:aws-deployment-framework-bootstrap/*_ multiple times in a short period of time, it leads to an error inside _CodeBuild:aws-deployment-framework-base-templates_.

# Info
* ADF version: 3.0.6
* Reproducible: Always

# Expected Behavior
CodeBuild should run multiple times without any issue. (1. run with the first CodeCommit change, 2. run with the second change, ...)

# Steps to reproduce
1. Edit any file inside _CodeCommit:aws-deployment-framework-bootstrap_ and push it.
2. Do an edit+push again (within 1-2 minutes)
3. Go to _CodeBuild:aws-deployment-framework-base-templates_ and you can see the error:

```
CLIENT_ERROR: NoSuchKey: The specified key does not exist. status code: 404,
request id: 2C756B0562D8D657,
host id: MY+Nw7YKRaPmZpHizWSx0P5F7luKZXChIccEOvBxWQHgDWGvCSKy7/ZOhwQIVuEiETzDk2qN3X0=
for primary source and source version
arn:aws:s3:::serverlessrepo-aws-deplo-bootstraptemplatesbucket-[...]/aws-deployment-frame/TemplateSo/[...]
```
![image](https://user-images.githubusercontent.com/1017301/83844728-cf22de80-a707-11ea-8b90-1c9d9ee53619.png)

# Reason
* ADF itself re-uses the CodePipeline/CodeBuild S3 artifact bucket
    1. [as Artifact bucket](https://github.com/awslabs/aws-deployment-framework/blob/3.0.6/src/template.yml#L499)
    2. [for saving data created by CodeBuild](https://github.com/awslabs/aws-deployment-framework/blob/3.0.6/src/template.yml#L458).
* Additionally [CodeBuild will remove all of the CodePipeline Artifacts](https://github.com/awslabs/aws-deployment-framework/blob/3.0.6/src/template.yml#L487)

# Solution
Use a separate ArtifactBucket for CodePipeline and for CodeBuild to store its created data.

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
[...]

PR #350, opened 2021-04-12
Lock down buckets created by ADF - block public access

sbkok:
**Why?**

By default, ADF does not mark any object or bucket to be accessibly
publicly. However, the buckets did not restrict any other process from
doing so anyway.

**What?**

To ensure objects are not accidentally made public, this change will
enforce that. Changing the object or bucket policies to public would not
be allowed.

Additionally, one bucket, the `BootstrapArtifactStorageBucket` that was
added recently by #270 did not configure encryption and versioning.
This was not released as part of ADF yet, so no data was exposed.

The samples that deployed buckets are also updated to include the
encryption, versioning, and block-public access properties.


deltagarrett (approved):
LGTM and... security is job zero! ; )

thomasmcgannon (approved):
LGTM, assuming line 778 with version is being replaced in #351 with the metadata tag?

sbkok:
@thomasmcgannon Correct, I will patch that in the other PR.

PR #351, opened 2021-04-12
Improve ADF version references in main template

sbkok:
**Why?**

For each new version of ADF, we need to update the version number in the
main template. However, as there are quite a few, it is likely to forget
one or search/replace another string that should not have matched.

**What?**

By adding the version number as a mapping, it can be updated quickly.
Reducing the number of times the version is repeated in the template to
two.

The first usage, in the Serverless Application Repository (SAR) metadata
is not able to use intrinsic CloudFormation functions unfortunately.
As described in the [SAR Metadata property documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-template-publishing-applications-metadata-properties.html).


deltagarrett (approved):
LGTM : )

thomasmcgannon (dismissed):
LGTM!

thomasmcgannon (approved):

deltagarrett (approved):
LGTM

PR #352, opened 2021-04-12
Add missing requirements file for shared python helpers

sbkok:
**Why?**

The `retrieve_organization_accounts.py` helper is using external
libraries that need to be installed for it to operate.

It seems the `requirements.txt` file lost its way when the retrieve
organization accounts helper was added by pull request #204.

**What?**

No need to worry, this change set adds the missing helper file.
The external library versions used match with the versions used in
the other requirement files in ADF.


deltagarrett (approved):
LGTM

thomasmcgannon (approved):
LGTM

Issue #337, opened 2021-03-12
Question: Disable webhook for Github provider


engrun:
We have a case where we need to create pipelines that are not triggered automatically by new commits to GH-repo. 

Looking at the docs, there doesn't seem to an property that disables the webhook creation or in any other way disables triggering of pipelines

Does anybody know it this is at all possible?

Looking at possible workarounds. 
1. Remove webhooks from GH, but I guess these will be re-applied when the "pipelines-pipeline" is re-run?
2. Write a script the de-registers the webhook from the pipeline. definitely will be re-applied when the "pipelines-pipeline" is re-run

3. Create a job that runs at periodic intervals that does either 1 og 2

Thoughts from the community?

thomasmcgannon:
Hi Rune,
In your use case, how would you expect the pipeline to be triggered, would this be through manually running the pipeline?
Currently there is no way to stop a pipeline from executing when there is a change in the source, however a similar feature (#308) was also requested.

Would having an option in your pipeline definition to not execute automatically solve your problem?

benbridts:
I haven't tested this, but you could solve part of this by adding an approval before the first deploy (the build would still run though)

engrun:
@thomasmcgannon 
thanks for the pointer to the other issue. This is in fact one of our cases. The other being manually triggering the pipeline as you say. 

And yes, and option for this in the yaml would be perfect. 👍 
eg. 
```
Enabled : True | False
or 
Trigger_on_update : True | False
```

sbkok:
Thank you for your patience. I am happy to inform you that this feature has been [released as part of v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues related to this feature.

Issue #308, opened 2020-12-08
Starting pipeline only on completion_trigger


pszemko-t:
Hi. 
A question or feature request... TBH I have no idea how to set this up currently.
I have a config for two pipelines and the second needs to be started only if the first one finishes 
In the current setup when I upload the file to 
          bucket_name: bucket.name
          object_key: key_2/package.zip
the second pipeline starts. 
But how to configure it in a way that second will run ONLY after it was triggered from first pipeline completion_trigger ?

```
pipelines:
# prerequisites pipeline
  - name: first-pipeline
    regions: eu-west-1
    default_providers: &generic_providers
      source: &generic_source
        provider: s3
        properties:
          account_id: 123412341234
          bucket_name: bucket.name
          object_key: key_1/package.zip
      build:
        provider: codebuild
        properties:
          environment_variables:
            CONTAINS_TRANSFORM: True
      deploy:
        provider: cloudformation
        properties:
          action: REPLACE_ON_FAILURE
          stack_name: adf-something-prereq
    params: 
      restart_execution_on_update: True
    targets: &generic_config_targets
      - path: /deployment
        name: Deploy-to-deployment
    completion_trigger:
      pipelines:
        - config-eu-west-1

  - name: second-pipeline
    regions: eu-west-1
    default_providers: &generic_config_providers
      source:
        provider: s3
        properties:
          account_id: 123412341234
          bucket_name: bucket.name
          object_key: key_2/package.zip
      build:
        provider: codebuild
        properties:
          environment_variables:
            CONTAINS_TRANSFORM: True
      deploy:
        provider: cloudformation
        properties:
          action: REPLACE_ON_FAILURE
          stack_name: adf-something
    params:
      restart_execution_on_update: True
    targets: *generic_config_targets
```

sbkok:
Hi @cloudedtales,

Unfortunately this is not supported at the moment.
I had a look at the source code, it generates the Source configuration for S3 [here](https://github.com/awslabs/aws-deployment-framework/blob/31f4719f83e80bd165b1567fed27de8eff090460/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codepipeline.py#L67-L68).
That would need to set the `PollForSourceChanges` property to `False` in order to stop CodePipeline from polling the target bucket/object.

We don't support this for the other source providers either as far as I am aware.
This would be a great first PR to add support for this, as the complexity would be low.
[...]

PR #357, opened 2021-04-21
Add ability to disable trigger on changes for S3/CC/GH source providers

sbkok:
## Why?

For several use cases, it would be great to have a pipeline that would not trigger on source updates.
For example, a pipeline might only need to be triggered on a schedule, by the completion of another pipeline, or manually.

This was requested by the following issues:
* Issue #308 - Starting pipeline only on completion_trigger
* Issue #337 - Question: Disable webhook for Github provider

## What?

The CodeCommit, Github, and S3 source providers have been update to support disabling triggering on source changes.

As described in the `docs/providers-guide.md` documentation, you can optionally disable it when required.

This code also solves a bug when a pipeline would rely on polling of changes for CodeCommit. In the previous version, it would be triggering twice instead of once when the source account id was specified. As the event handler would be listening for events that indicate an update was performed, even when the pipeline was set to use polling for changes instead of events.

---

### DIFF
--- a/docs/providers-guide.md
+++ b/docs/providers-guide.md
@@ -76,6 +76,17 @@ Provider type: `codecommit`.
   > The role to use to fetch the contents of the CodeCommit repository.
   > Only specify when you need a specific role to access it. By default ADF
   > will use its own role to access it instead.
+- *trigger_on_changes* - *(Boolean)* default: `True`.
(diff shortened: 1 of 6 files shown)


thomasmcgannon (approved):
Thanks for implementing this feature!

deltagarrett (approved):
LGTM, great stuff!

PR #360, opened 2021-05-12
GitHub default repository name change from master to main

ippetrus:
*Issue #, if available:* None

*Description of changes:*
Github changed the name of the default branch name which is created when you create a new repository from **master** to **main**. There are two minor changes with this pull request. Only the default github branch is changed from master to main.
Also this name change is updated in the doc file too.

### DIFF
--- a/docs/providers-guide.md
+++ b/docs/providers-guide.md
@@ -101,7 +101,7 @@ Provider type: `github`.
   > The GitHub repository name.
   > For example, for the ADF repository it would be:
   > `aws-deployment-framework`.
-- *branch* - *(String)* - default: `master`.
+- *branch* - *(String)* - default: `main`.
   > The Branch on the GitHub repository to use to trigger this specific
   > pipeline.
 - *owner* - *(String)* **(required)**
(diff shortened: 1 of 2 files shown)


sbkok (changes requested):
Thanks for proposing this change @ippetrus!
Would be great to have support for this and rely on main as the default branch.

The current change you proposed would imply breaking change to every user that has GitHub setup but still relies on the master branch. Could you make it a configuration property as part of the adfconfig.yml, which would:

- When installed as new, use `main` in the adfconfig as the default.
- When upgrading and no config is present, would use `master` in the adfconfig as the default.

Plus it would be great to show a deprecation warning for customers that are upgrading and relying on the master branch as being the default. In the future, as part of a v4.0 release, I would like to phase out the term master all together and rely on `main` as the default branch for all installs/upgrades.

If you don't know how to start on this, please reach out, I'm happy to help out getting this sorted. This would be a great improvement to ADF I think.

sbkok:
I'm closing this PR, as it is superseded by #370.

PR #362, opened 2021-05-17
Fix duplicate notification endpoint setup in pipeline generation

sbkok:
**Why?**

When generating a pipeline, the code optimization introduced by PR #285
broke the notification_endpoint configuration. When set, it would try to
configure the notification in the main generator, as well as in the
default ADF pipeline generator.

**What?**

Removing the notification endpoint configuration from the main template,
as that is replaced by the default ADF pipeline generator
implementation.


deltagarrett (approved):
LGTM

PR #370, opened 2021-07-02
Feature: added ability to override the default branch

apogorielov:
…pipelines

*Issue #, if available:*

*Description of changes:*

Following PR adds ability to override default "master" branch for deployment pipelines using values provided in adfconfig.yml. If parameter is not provided, branch defaults to "master" to remain backward compatibility.  

Priority of branch selection:
1. Deployment map branch configuration parameter.
2. Adfconfig default branch parameter.
3. Master branch. 

Parameter is propagated to the deployment accounts using SSM, where later is retrieved and used a part of input generation for pipelines. 


src/lambda_codebase/initial_commit/bootstrap_repository/example-adfconfig.yml
```diff
@@ -15,5 +15,6 @@ config:
       action: safe  # Can be safe or remove-base (see docs)
   scp:
     keep-default-scp: enabled  # determines if the default AWSFullAccess SCP stays attached to all OU's 
-  scm: # Source control management, if true and using codecommit as source the repository will be automatically created
-    auto-create-repositories: enabled
\ No newline at end of file
+  scm: # Source control management
+    auto-create-repositories: enabled # if true and using codecommit as source the repository will be automatically created
+    default-scm-branch: master        # allow to specify custom defualt branch if pipeline specific is not provided. If parameter is not specified -  "master" branch will be used.
```
sbkok, line 20:
```suggestion
    default-scm-branch: master        # allow to specify custom default branch if a pipeline specific configuration is not provided. If the parameter is not specified it defaults to the "master" branch.
```

sbkok (changes requested):
Small change requested.
Other than that, looks good to me.

I wish we could default to main as the branch name. But unfortunately that would not be backward compatible. Something to consider for the next major release.

Thanks for contributing this change.

src/lambda_codebase/initial_commit/bootstrap_repository/example-adfconfig.yml
sbkok, line 20: Could you also cover these settings in the documentation for administrators?

sbkok (approved):
I've pushed the changes requested in the review comments above.
Thanks for delivering this feature @apogorielov!

StewartW (approved):
Looks good to me :shipit:

Issue #400, opened 2021-11-05
Timeout on Deployment Account's Step Function


stemons:
Hello team, we have an issue on Deployment Account's Step Function.
Every time the user push some changes to bootstrap repository or just release change, the step function in deployment account reach the 900 seconds timeout.
The environment consists of about 200 accounts and 18 regions.
Step function is trying to iterate over all the accounts and regions to apply the changes to adf-cloudformation-role (adding regional S3 bucket and kms to the policy) even if the policy is correct and no need additional changes.
Any workaround to skip the step function if not needed and proceed with the other bootstrap activities?

sbkok:
Thanks for reporting the issue.
As you might have noticed, @StewartW is working on a fix for this issue.

Once merged, I will leave this issue open until it is released as part of the next version of ADF.

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
Please open a new issue if you are experiencing any issues with the latest release.

PR #401, opened 2021-11-08
Fix: Deployment Account Step Function Time outs

StewartW:
*Issue #, if available:* #400 

*Description of changes:*
Refactored state machine slightly to move account iteration logic into
a map, lambda function then does region checks for each account as before.

No change to function logic bar removing the for account_id in account_ids
loop.


src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/enable_cross_account_access.py
```diff
@@ -36,6 +36,7 @@ def update_iam(role, s3_bucket, kms_key_arn, role_policies):
 
 
 def lambda_handler(event, _):
+    LOGGER.debug(event)
```
sbkok, line 39: Can you remove this debug line, or only expose the data that really needs to be present + explaining where we are at / what we are processing?

Like
```bash
"Setting up cross-account access for ${account_id} in ${regions}"
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
```diff
@@ -1118,12 +1118,29 @@ Resources:
...
                     "Default": "EnableCrossAccountAccess"
                 },
-                "EnableCrossAccountAccess": {
-                    "Type": "Task",
-                    "Resource": "${EnableCrossAccountAccess.Arn}",
-                    "Next": "UpdateDeploymentPipelines",
-                    "TimeoutSeconds": 900
-                },
+                "EnableCrossAccountAccessMap": {
+                      "Type": "Map",
```
sbkok, line 1122: The indentation is off, could you fix that?
Probably happened when the partition PR got merged.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
```diff
@@ -1118,12 +1118,29 @@ Resources:
...
+                      },
+                      "ItemsPath": "$.account_ids",
+                      "Parameters": {
+                        "deployment_account_region.$": "$.deployment_account_region",
+                        "deployment_account_id.$": "$.deployment_account_id",
+                        "account_id.$": "$$.Map.Item.Value",
+                        "regions.$": "$.regions"
+                      },
+                      "ResultPath": null
+                    }
```
sbkok, line 1143: I think there is missing comma here:
```suggestion
                    },
```

sbkok: Oddly enough, the next line indents on a slightly different line it seems. Can you check this?

sbkok (changes requested):
Thanks for working on this issue.
Could you look at the attached comments?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/enable_cross_account_access.py
StewartW: I'll just get rid of it. It's a hang over from debugging the step function map.

sbkok (approved):
LGTM, thank you!

dsudduth (approved):
The latest round of changes look great! Thank you for the fix!

Issue #379, opened 2021-08-03
Are IAM Permissions for Support API operations missing for CodeBuild role?


rickardl:
Hi,

We were previously on another internal version of ADF, as we moved to `3.1.2` we saw a failure of being able to use the `support_level` integration, am I correct to assume it was never added in #232, because I don't see a reference to any changes in the template for ADF granting permissions.

It causes an issue
![image](https://user-images.githubusercontent.com/749326/127995739-ba326e17-55e5-45b0-9221-6656206150f7.png)

Adding the rights to the computed `adf-codebuild-role` role resolves the issue.

```diff
--- a/src/template.yml
+++ b/src/template.yml
@@ -407,6 +407,7 @@ Resources:
               - 'sts:GetCallerIdentity'
               - 'sts:assumeRole'
               - 'cloudformation:ValidateTemplate'
+              - 'support:*'
             Resource: '*'
           - Effect: 'Allow'
             Action:
```

If this is correct, would you accept adding 'support:*' or do you feel more comfortable with a narrower permission only granting the needed rights or using a Managed Policy? Let me know :)

sbkok:
Hi @rickardl,

Thanks for reporting this.
Adding it to the `adf-codebuild-role` would enable any pipeline that uses that role (which is the default role) to perform similar actions. I will work on a pull-request that will address this shortcoming, without granting the default role this support permission. Using restricted actions.

I hope to have an implementation ready next week.
So we can put it in the next release and you can copy the changes to your to local ADF installation to resolve it securely.

Best regards,
Simon

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
Please open a new issue if you are experiencing any issues with the latest release.

PR #402, opened 2021-11-08
Fix permission to set Support Subscription upon account creation

sbkok:
**Why?**

As reported in #379, the IAM permission required to enable Support Subscriptions upon account creation is missing.

**What?**

This change adds another Role that is used by the Bootstrap CodeBuild project to request the right support level upon account creation.

A separate role is required, as the permissions should not be added to the default ADF CodeBuild role that is also used by the pipelines generated by ADF.

---


StewartW (approved):

deltagarrett (approved):
LGTM

Issue #380, opened 2021-08-06
Feature: possibility to have more than one SCP in adf-bootstrap


GiuDB:
Hi,

when bootstrapping accounts via ADF it's possible to use "only" one Service Control Policy (in the form of scp.json file within the adf-bootstrap folder from the aws-deployment-framework-bootstrap repository on the Master Account).

When such scp.json exceed the maximum allowed limit of 5KB, it cause an issue:
`An error occurred (ConstraintViolationException) when calling the UpdatePolicy operation: You have exceeded the maximum policy size`

It would be good to have adf managing more than one SCPs, like for example in a folder like this:
scp
|__ scp1.json
|__ scp2.json
|__ ...

PR #404, opened 2021-11-09
Adding support for multiple policies per target.

srabidoux:
*Issue #, if available:* 380 (https://github.com/awslabs/aws-deployment-framework/issues/380)

*Description of changes:*

The following changes allow users to attach multiple policies to a single target using the bootstrap process. Any files with name `scp*.json` (or `tagging-policy*.json`, for tagging policies) will be picked up and applied. The naming convention used for the policies is

'adf-{0}-{1}--{2}'.format(_type, path, policy_filename)

for policies named anything other than '{0}.json'.format(_type), but, for backwards-compatibility, policies that come from files named exactly '{0}.json'.format(_type) are named 

'adf-{0}-{1}'.format(_type, path)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -35,12 +35,9 @@ def _compare_ordered_policy(self, obj):
...
     @staticmethod
-    def _trim_scp_file_name(policy):
-        return policy[1:][:-8] if policy[1:][:-8] == '/' else policy[2:][:-9]
-
-    @staticmethod
-    def _trim_tagging_policy_file_name(policy):
-        return policy[1:][:-19] if policy[1:][:-19] == '/' else policy[2:][:-20]
+    def _trim_policy_file_name(policy):
+        # Returns policy target and filename as tuple
+        return '/' if '/'.join(policy.split('/')[1:-1]) == '' else '/'.join(policy.split('/')[1:-1]), policy.split('/')[-1]
```
sbkok, line 40: This code looks to complex for what it is doing I am afraid.

Could you change this function to use a normal if, instead of an inline one.
And could you use variables and add comments to explain what is happening exactly?

It would be good to add tests for this too, inside the `src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/tests/` directory. Let me know if you need help with that, as this file is not tested at all I see.
I'm trying to raise our test coverage, as debugging/testing all use cases of ADF is extremely hard to accomplish otherwise.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -71,10 +68,13 @@ def set_scp_attachment(
     def clean_and_remove_policy_attachment(
             organization_mapping,
             path,
+            policy_filename,
             organizations,
             policy_type):
+        _type = 'scp' if policy_type == "SERVICE_CONTROL_POLICY" else 'tagging-policy'
+        policy_name = 'adf-{0}-{1}--{2}'.format(_type, path, policy_filename) if policy_filename != '{}.json'.format(_type) else 'adf-{0}-{1}'.format(_type, path)
```
sbkok, line 75: Could you make this line a function that returns the name to use?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -71,10 +68,13 @@ def set_scp_attachment(
     def clean_and_remove_policy_attachment(
             organization_mapping,
             path,
+            policy_filename,
             organizations,
             policy_type):
+        _type = 'scp' if policy_type == "SERVICE_CONTROL_POLICY" else 'tagging-policy'
```
sbkok, line 74: Would be good to change these hard coded strings into STATIC variables at the top.
Could you introduce this change?

sbkok: In the `apply` function, the variable `_type` uses the exact opposite, setting it to either `'SERVICE_CONTROL_POLICY' or 'TAG_POLICY'`. It would be easier to keep this consistent, and only tansition to the short version when we need it in an output.

So in other words, I would suggest to use the longer version all the time in the code. And only when we need to output it or set it as part of a path convert it to the print/path friendly version through calling a function that transforms it to that style.

If you change the next line into a function, you only need to pass the `policy_type` for this function.

sbkok: Could you change L100 to that style too? The one with `for policy in ['scp', 'tagging-policy']:`.
To something like:

```
for policy_type in [SERVICE_CONTROL_POLICY, TAGGING_POLICY]:
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -116,33 +116,34 @@ def apply(self, organizations, parameter_store, config):  # pylint: disable=R091
...
                     if stored_policy not in _policies:
                         OrganizationPolicy.clean_and_remove_policy_attachment(
-                            organization_mapping, path, organizations, _type)
+                            organization_mapping, path, policy_filename, organizations, _type)
             except ParameterNotFoundError:
                 LOGGER.debug(
                     'Parameter %s was not found in Parameter Store, continuing.', policy)
                 pass
 
             for _policy in _policies:
```
sbkok, line 125: Could you change the name of the variable `_policy` to `policy_path`. And change `_policies` to `policy_paths`.
I don't know why these are using underscores in front, as they are not class/instance variables that need protection, would be good to revert that. I hope you don't mind making these small improvements to the file on the side 🤞. If you would rather not, I am happy to push another PR for that later too.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -116,33 +116,34 @@ def apply(self, organizations, parameter_store, config):  # pylint: disable=R091
...
-                path = OrganizationPolicy._trim_scp_file_name(
-                    _policy) if policy == 'scp' else OrganizationPolicy._trim_tagging_policy_file_name(_policy)
+                path, policy_filename = OrganizationPolicy._trim_policy_file_name(
+                    _policy)
+                policy_name = 'adf-{0}-{1}--{2}'.format(policy, path, policy_filename) if policy_filename != '{}.json'.format(policy) else 'adf-{0}-{1}'.format(policy, path)
```
sbkok, line 128: This would become a function call again.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -20,7 +20,7 @@ def __init__(self):
     @staticmethod
     def _find_all(policy):
         _files = list(glob.iglob(
-            './adf-bootstrap/**/{0}.json'.format(policy),
```
sbkok, line 23: Could you rename this function to _find_all_paths, to clarify what it will return.
As this is part of the `OrganizationPolicy` class, one could otherwise think it is returning instances of that class.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -70,17 +70,11 @@ def update_policy(self, content, policy_id):
             Content=content
         )
 
-    def create_policy(self, content, ou_path, policy_type="SERVICE_CONTROL_POLICY"):
+    def create_policy(self, content, ou_path, policy_name, policy_type="SERVICE_CONTROL_POLICY"):
```
sbkok, line 73: The `ou_path` is no longer used in this function. Could you remove that?

sbkok (changes requested):
Thanks for adding this!

I left a few comments in the files that were changed.
Could you also add documentation on how to use this feature, so our users know how to perform the change? I think the admin guide would make most sense for this.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
srabidoux: Done.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
srabidoux: Good idea. Fixed.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
srabidoux: Fixed in recent commit.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
srabidoux: Made these changes.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
srabidoux: I wasn't quite sure what you meant by "Would be good to change these hard coded strings into STATIC variables at the top.", but I made the rest of the suggested changes in the recent commit.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
srabidoux: Yes, done.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
srabidoux: I simplified/commented this routine.

As far as tests go, I may need some help getting that started for this file. If you could do get a basic test file for organization_policy.py started, I can help expand upon that.

srabidoux:
@sbkok I made most of the suggested changes, and left comments under the issues that still remain for your feedback. The admin-guide.md has been updated as well.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
```diff
@@ -67,14 +75,25 @@ def set_scp_attachment(
...
                     pass
 
+    @staticmethod
+    def return_policy_name(policy_type, target_path, policy_filename):
+        _type = 'scp' if policy_type == "SERVICE_CONTROL_POLICY" else 'tagging-policy'
+        if policy_filename != '{}.json':
+            return 'adf-{0}-{1}--{2}'.format(_type, target_path, policy_filename)
+        else:
+            # Added for backwards-compatibility with previous versions of ADF
+            return 'adf-{0}-{1}'.format(_type, target_path)
```
sbkok, line 85: The file name is not inserted into the string at the moment.
I think it needs to be changed to an if against a format string.

Additionally, it will not continue code execution after a return, so we can shorten it too by removing the `else`.

Considering the longer string in the if statement, I would propose we flip those, so we skip indenting the longest line.

All together, you would have something like:

```suggestion
        if policy_filename == f'{_type}.json':
            # Added for backwards-compatibility with previous versions of ADF
            return 'adf-{0}-{1}'.format(_type, target_path)
        return 'adf-{0}-{1}--{2}'.format(_type, target_path, policy_filename)
```

sbkok (changes requested):
Thank you for making these changes.
I've attached two more review comments here.

I will add a separate commit with a test example shortly.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/organization_policy.py
StewartW: I've added in a basic test that tests the policy name function.

sbkok:
@srabidoux I see you managed to write a few test in the meanwhile.
I assume you don't need my help with writing those any more in that case.

If I assumed wrong, please ping me here and I will try to help with those.

sbkok:
I'm closing this pull request for now, because it has been inactive for too long.

Feel free to reopen it though :) In case you want to reopen, could you update the pull request, rebase your branch, and/or clarify the changes made? Please let us know if you need help with this.

PR #414, opened 2021-12-23
Fix duplicated steps in Account Bootstrap Step Function

sbkok:
### Why?

In one of our recent PRs the indentation of the state machine was
changed. However, with this update, it seems that part of the state
machine got copied.

Because of this, Step Function was unable to deploy/update the state
machine.

### What is changed?

Removed the redundant parts of the state machine.
I checked both definitions, the repeated segments are a 1:1 copy.

---

### DIFF
--- a/src/template.yml
+++ b/src/template.yml
@@ -772,79 +772,24 @@ Resources:
                         "Choices": [{
                             "Variable": "$.is_deployment_account",
                             "NumericEquals": 1,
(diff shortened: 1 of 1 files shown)


deltagarrett (approved):
LGTM

krisnielander (approved):
Looking better, approved 👍

PR #415, opened 2021-12-24
Upgrade to Python 3.9

sbkok:
**Why?**

To enable use of features added to Python 3.9.

**What?**

Updated the CodeBuild container references, Lambda execution
environment, and samples to rely on Python 3.9.

---


deltagarrett (dismissed):
LGTM

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
```diff
@@ -57,6 +63,7 @@ Resources:
       CompatibleRuntimes:
         - python3.7
         - python3.8
+        - python3.9
```
Nr18, line 66: I am going to do an assumption here so it might be incorrect. 

The reasoning behind the PR was the ability to utilize the new advantages that 3.9 has to offer. But the shared layer stays compatible with 3.7 and 3.8. 

Therefore you cannot use the new features of 3.9 or you need build a compatible equivalent for older versions.

Is that something you have considered? Is there a specific reason for keeping the shared later compatible with other versions?

sbkok: Good point, thanks for catch that.
I will remove the older Python versions from that list.

StewartW (approved):
Looks good to me. 
One question about the sample buildspecs being updated. 
Will customers have to do this in their own repositories or is it just over zealous find and replace?

sbkok:
About the `samples/**/buildspec.yml`: Customers don't have to update those locally, but they can if they want to.

I just want to help any new ADF user that looks for copy/paste-able examples to use an up-to-date base.

thomasmcgannon (approved):
lgtm!

PR #418, opened 2021-12-31
Update docs to state the default branch used as the source

sbkok:
**Why?**

The docs stated that the default branch would always default to `master`
when the branch is not specified.

However, the branch that is used is determined by looking up the
`config/scm/default-scm-branch` as configured in the `adfconfig.yml` file.

**What?**

Updated the docs accordingly.

---


StewartW (approved):

thomasmcgannon (approved):
lgtm!

PR #421, opened 2022-01-05
Fix global-iam example comment explaining where it is deployed

sbkok:
**Why?**

At the moment, the `global-iam.yml` is stating that it is deployed to the deployment account. However, this specific file is deployed to all target accounts instead.

**What?**

Simple, updated the comment.

---


thomasmcgannon (approved):
lgtm!

StewartW (approved):

PR #424, opened 2022-01-10
Refactor:  Separate Pipeline Management into Serverless Application

StewartW:
*Issue #, if available:*

*Description of changes:*


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/deployment_map.py
```diff
@@ -101,4 +100,4 @@ def _process_dir(self, path):
                     self._read(filename)
                 )
             else:
-                LOGGER.warning("%s is not a directory and doesn't end in.yml", filename)
+                LOGGER.warning(f"{filename} is not a directory and doesn't end in.yml")
```
sbkok, line 103: Could you please revert this change? The logger has a capability to insert strings only when the logging level is enabled. So the filename will not be inserted if logging is disabled. In this case probably all loggers would pick up on warnings. However, I would like to keep those consistent if possible.

```suggestion
                LOGGER.warning("%s is not a directory and doesn't hold the .yml suffix", filename)
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/repo.py
```diff
@@ -23,8 +22,6 @@
     DEPLOYMENT_ACCOUNT_REGION,
     S3_BUCKET_NAME
 )
-
-
 class Repo:
```
sbkok, line 25: The lines that were removed before the class should be there.
Adding these as a suggestion here, as GitHub doesn't support suggestions on deleted lines (odd).
```suggestion


class Repo:
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/rule.py
```diff
@@ -54,5 +54,5 @@ def create_update(self):
             s3_key_path=None,
             account_id=DEPLOYMENT_ACCOUNT_ID,
         )
-        LOGGER.info('Ensuring Stack State for Event Rule forwarding from %s to %s', self.source_account_id, DEPLOYMENT_ACCOUNT_ID)
+        LOGGER.info(f'Ensuring Stack State for Event Rule forwarding from {self.source_account_id} to {DEPLOYMENT_ACCOUNT_ID}')
```
sbkok, line 57: Same comment as above on the logger and use of `%s` being preferred over formatted strings with loggers:
```suggestion
        LOGGER.info('Ensuring Stack State for Event Rule forwarding from %s to %s', self.source_account_id, DEPLOYMENT_ACCOUNT_ID)
```

src/template.yml
```diff
@@ -17,12 +17,12 @@ Metadata:
     Labels:
       ["adf", "aws-deployment-framework", "multi-account", "cicd", "devops"]
     HomePageUrl: https://github.com/awslabs/aws-deployment-framework
-    SemanticVersion: 3.1.2
+    SemanticVersion: 3.4.1
```
sbkok, line 20: This should not be changed in this PR:
```suggestion
    SemanticVersion: 3.1.2
```

src/template.yml
```diff
@@ -17,12 +17,12 @@ Metadata:
...
     SourceCodeUrl: https://github.com/awslabs/aws-deployment-framework
 Mappings:
   Metadata:
     ADF:
-      Version: 3.1.2
+      Version: 3.4.1
```
sbkok, line 25: This should not be changed in this PR:

```suggestion
      Version: 3.1.2
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -48,13 +48,13 @@ def _define_target_type(target):
...
     def generate_waves(self):
-        waves = []
         wave_size = self.wave.get('size', 50)
+        wave = []
         length = len(self.account_list)
         for index in range(0, length, wave_size):
-            yield self.account_list[index:min(index + wave_size, length)]
-            waves.append(self.account_list[index:min(index + wave_size, length)])
-        return waves
+            wave.append(self.account_list[index:min(index + wave_size, length)])
```
StewartW, line 55: I've changed this back as it had issues serializing the response due to a generator being used.

src/template.yml
StewartW: Woops, changed this for deploying locally

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/cloudformation.py
```diff
@@ -178,21 +180,21 @@ def _create_change_set(self):
...
-                    ChangeSetType=self._get_change_set_type())
+                change_set_params = {
+                    "StackName": self.stack_name,
+                    "TemplateURL": self.template_url,
+                    "Parameters": self.parameters if self.parameters is not None else self.get_parameters(),
+                    "Capabilities": ["CAPABILITY_NAMED_IAM", "CAPABILITY_AUTO_EXPAND"],
+                    "Tags":[{
+                        'Key': 'createdBy',
+                        'Value': 'ADF'
+                        }],
```
sbkok, line 191:
```suggestion
                    }],
```

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,879 @@
...
+                  - cloudformation:ExecuteChangeSet
+                  - cloudformation:SetStackPolicy
+                  - cloudformation:ValidateTemplate
+                Resource:
+                  - "*"
+              - Effect: Allow
+                Sid: "PassRole"
+                Action:
+                  - 'iam:PassRole'
+                Resource: '*'
```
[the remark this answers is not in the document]
StewartW: I think that can be removed. (It's a carry over.) Can you just double check. I think the CodeBuild Project role will never need to  pass an IAM role

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,879 @@
...
+                  - "iam:CreateRole"
+                  - "iam:CreateRolePolicy"
+                  - "iam:DeleteRole"
+                  - "iam:DeleteRolePolicy"
+                  - "iam:GetRole"
+                  - "iam:GetRolePolicy"
+                  - "iam:PassRole"
+                  - "iam:PutRolePolicy"
+                Resource:
+                  - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:role/*
```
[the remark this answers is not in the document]
StewartW: I think this can be removed as well. Based on my reasoning this role should only be used to deploy CodePipeline stacks which don't include roles. 

Think this is just hangover from previous iterations.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+      Description: "ADF Lambda Function - Create Repository"
+      Environment:
+        Variables:
+          ACCOUNT_ID: !Ref AWS::AccountId
+          ORGANIZATION_ID: !Ref OrganizationID
+          ADF_VERSION: !Ref ADFVersion
+          ADF_LOG_LEVEL: !Ref ADFLogLevel
+          ADF_ROLE_NAME: !Ref CrossAccountAccessRole
+          S3_BUCKET_NAME: !Ref PipelineBucket
+      FunctionName: CreateRepositoryFunction
```
sbkok, line 777:
```suggestion
      FunctionName: ADFPipelineCreateRepositoryFunction
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+      Environment:
+        Variables:
+          ACCOUNT_ID: !Ref AWS::AccountId
+          ORGANIZATION_ID: !Ref OrganizationID
+          ADF_VERSION: !Ref ADFVersion
+          ADF_LOG_LEVEL: !Ref ADFLogLevel
+          ADF_ROLE_NAME: !Ref CrossAccountAccessRole
+          S3_BUCKET_NAME: !Ref PipelineBucket
+          ROOT_ACCOUNT_ID: !Ref RootAccountId
+      FunctionName: GeneratePipelineInputsFunction
```
sbkok, line 794:
```suggestion
      FunctionName: ADFPipelineGenerateInputsFunction
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+      Environment:
+        Variables:
+          ACCOUNT_ID: !Ref AWS::AccountId
+          ORGANIZATION_ID: !Ref OrganizationID
+          ADF_VERSION: !Ref ADFVersion
+          ADF_LOG_LEVEL: !Ref ADFLogLevel
+          ADF_ROLE_NAME: !Ref CrossAccountAccessRole
+          S3_BUCKET_NAME: !Ref ADFDefinitionBucket
+          ROOT_ACCOUNT_ID: !Ref RootAccountId
+      FunctionName: StoreDefinitionFunction
```
sbkok, line 811:
```suggestion
      FunctionName: ADFPipelineStoreDefinitionFunction
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+        Variables:
+          ACCOUNT_ID: !Ref AWS::AccountId
+          ORGANIZATION_ID: !Ref OrganizationID
+          ADF_VERSION: !Ref ADFVersion
+          ADF_LOG_LEVEL: !Ref ADFLogLevel
+          ADF_ROLE_NAME: !Ref CrossAccountAccessRole
+          ROOT_ACCOUNT_ID: !Ref RootAccountId
+          S3_BUCKET_NAME: !Ref ADFPipelineBucket
+          ADF_PIPELINE_PREFIX: !Ref PipelinePrefix
+      FunctionName: IdentifyOutOfDatePipelinesFunction
```
sbkok, line 829:
```suggestion
      FunctionName: ADFPipelineIdentifyOutOfDatePipelinesFunction
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/deployment_map.py
```diff
@@ -83,10 +92,8 @@ def _get_all(self):
         )
         if not self.map_contents['pipelines']:
             LOGGER.error(
-                "No Deployment Map files found, create a deployment_map.yml "
-                "file in the root of the repository to create pipelines. "
-                "You can create additional deployment maps if required in a "
-                "folder named deployment_maps with any name (ending in .yml)"
+                "No Deployment Map files found, create a deployment_map.yml file in the root of the repository to create pipelines. "
+                "You can create additional deployment maps if required in a folder named deployment_maps with any name (ending in .yml)"
```
sbkok, line 96: Can you revert this change, so the line length matches what we defined in the .editorconfig file?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/test_deployment_map.py
```diff
@@ -0,0 +1,68 @@
...
+                        "regions": ["eu-west-1"],
+                        "step_name": "",
+                    },
+                ]
+            ]
+        ]
+    }
+
+    cls.update_deployment_parameters(pipeline)
+    assert cls.account_ou_names["some_account"] == "/fake/path"
```
sbkok, line 68: Could you add the tests I added in the #484 Fix: Wave deployments PR. That way I can close that one in favour of this one :smile:

See https://github.com/awslabs/aws-deployment-framework/pull/484/commits/18a0492b46ade867baa8f1e7256d72bbbf92f93b

sbkok (changes requested):
Great improvements, thanks for making those.
I've got a number of minor suggested changes, so we are close to merge this too!

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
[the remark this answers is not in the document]
sbkok: I think it uses the PassRole when executing the CloudFormation stacks.
According to the change you introduced, it will always use one role, see https://github.com/awslabs/aws-deployment-framework/pull/424/files/7f9d1c95cba423d9e20533436652b6d172399ed3#diff-de3040c3b70530a31a46f3a9ce5ffde48ecf03da5153a64929720b8e53c334e6R58

The role is created and configured in this stack too, see https://github.com/awslabs/aws-deployment-framework/pull/424/files/7f9d1c95cba423d9e20533436652b6d172399ed3#diff-dead502985c4f4a35238ba4a93fcedfabd8d013bd70e545003066d4a01afe956R500-R501

So I think we can restrict it to:

```suggestion
                Resource:
                  - !GetAtt ADFPipelineMangementCloudFormationRole.Arn
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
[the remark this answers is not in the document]
sbkok: Unfortunately it does create a role depending on the resources in the pipeline.
For example, for CodePipeline it needs to create a role that has access.

That role name starts with `adf-pipeline-`. So maybe we can restrict this to only accept that as the prefix. What do you think?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+                  - "sns:Subscribe"
+                  - "sns:SetTopicAttributes"
+                  - "sns:GetTopicAttributes"
+                  - "sns:TagResource"
+                Resource:
+                  - !Sub arn:${AWS::Partition}:sns:${AWS::Region}:${AWS::AccountId}:${PipelinePrefix}*
+              - Effect: Allow
+                Action:
+                  - "iam:CreateRole"
+                  - "iam:CreateRolePolicy"
```
sbkok, line 672: This action does not exist, do you mean AttachRolePolicy?

```suggestion
                  - "iam:AttachRolePolicy"
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+                  - "iam:DeleteRole"
+                  - "iam:DeleteRolePolicy"
+                  - "iam:GetRole"
+                  - "iam:GetRolePolicy"
+                  - "iam:PutRolePolicy"
+                Resource:
+                  - !Sub arn:${AWS::Partition}:iam::${AWS::AccountId}:role/aws-deployment-framework/*
```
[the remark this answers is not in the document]
sbkok: Maybe we can have a separate scope for pipeline roles. So a pipeline will not be granted access to the more permissive admin roles that ADF needs to generate resources. Or we define a separate path for those.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -0,0 +1,886 @@
...
+                Condition:
+                  StringEquals:
+                    aws:PrincipalOrgID: !Ref OrganizationID
```
sbkok, line 682: This condition is not supported on any of the IAM actions listed above.
This statement would work with PassRole, and we still need that I am afraid.
As CodePipeline, for example, gets passed the role that we create using the allowed actions defined in this statement.

sbkok (changes requested):

sbkok (approved):
Thanks for making all these changes @StewartW!
This will help speed up ADF a lot, excellent work!

javydekoning (approved):
This looks good to me! 

I suggest also storing the `.drawio` originals to the repo so the diagram can be easily updated when there are changes to the statemachine later.

sbkok:
For reference, the .drawio.png files that were added by the PR also include the source of the image.
You can edit them by opening the png file directly with the DrawIO tool.

PR #428, opened 2022-01-13
feat: enable use of codebuild reports

Nr18:
*Issue #, if available:* N/A

*Description of changes:*

Currently, you cannot use the reports functionality in the build phases of your pipelines. This has to do with the fact that the IAM Role that is executing the build has no permissions to do so. This prevents you from collecting reports using the CodeBuild Reports functionality. By adding the needed IAM actions to the CodeBuild Role you will be able to collect your reports for unit tests and coverage reports.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
@@ -280,6 +280,15 @@ Resources:
               - "ecr:PutImage"
             Resource:
               - "*"
+          - Effect: Allow
+            Action:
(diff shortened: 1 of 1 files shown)


StewartW (approved):
LGTM

sbkok:
Thanks for sharing this implementation. I see a need for this and would like to support that in an upcoming release.

However, we cannot merge the code as-is unfortunately.
As this will open up any CodeBuild projects deployed by ADF to overwrite reports generated by other CodeBuild projects.

I am looking into a fix to grant access on a project by project level.
I have created a separate issue for this, as the request will require more things to be rewritten.
I will close this PR, we can move our discussion on this topic to issue #434.

Issue #422, opened 2022-01-05
Use KMS encryption for ADF notifications


Nr18:
## Why

From a compliance point of view we need all SNS topics to be encrypted. The pipeline notifications topic is currently not using encryption.

## Proposal

As Werner Vogels once said: 

> Dance Like Nobody’s Watching. Encrypt Like Everyone Is

So we should:
- ~Use the AWS managed KMS key by default.~
- ~Make use of a Customer Managed KMS Key optional.~
- Use the default KMS key created by ADF.

## Feedback?

Do you know the year Werner used that quote on stage at re:Invent? I can't remember the year... 

But please contribute in this discussion!

sbkok:
Thank you for your patience. I am happy to inform you that this feature has been [released as part of v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues related to this feature.

PR #429, opened 2022-01-13
feat: encrypt sns topic using the ADF CMK KMS Key

Nr18:
*Issue #, if available:* #422

*Description of changes:*
Because ADF already has a KMS key that is used for encryption. It makes sense to use that key for the SNS notifications as well.


StewartW (dismissed):
Makes sense. 
Might be good to make it configurable going forward but I'm happy with that as is.

sbkok (changes requested):
Thanks for contributing this @Nr18.

Nr18:
Ah that makes sense! I applied your suggestion.

StewartW (approved):

PR #436, opened 2022-02-04
feat: allow adf-global-base-deployment parameter overwrites

Nr18:
*Issue #, if available:*

*Description of changes:*

We use a prefix in the CodeCommit repo name. So all our stacks are already prefixed with that prefix.

For this reason we changed the StackPrefix to an empty string in the global.yml for the deployment account. This could not be done without an customization, so we removed the `adf-` prefix in the template.

This pull request adds support for setting parameters in a JSON file. I an not sure about the format of the file itself. Or that this will work in this form. Lets use this pull request to figure out what the best way is.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global-params.json
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global-params.json
@@ -1,0 +1,8 @@
+{
+    "Parameters": [
+        {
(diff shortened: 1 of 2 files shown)


sbkok:
I appreciate your effort to dive into this solution and share it with the community for those who face a similar need.

Unfortunately we cannot accept this PR as part of ADF. The proposed change would modify the prefix for all existing customers too. The reason behind the fixed stack prefix is to enable SCPs that only grant ADF access to manage these.

Issue #445, opened 2022-03-08
param_overrides does not support more than 1 parameter (per input)


Nr18:
When you make use of the `param_overrides` functionality and need to pass more than 1 parameter per input the pipeline will fail to deploy. For example:

```yaml
pipelines:
  - name: my-pipeline
    default_providers:
      source:
        provider: codecommit
        properties:
          account_id: XXXXXXXXXXXX
      build:
        provider: codebuild
        properties:
          image: STANDARD_4_0
          environment_variables:
            CONTAINS_TRANSFORM: True
      deploy:
        provider: cloudformation
    targets:
      - name: FirstStack
        regions: eu-west-1
        target: XXXXXXXXXXXX
        properties:
          outputs: outputs
      - name: SecondStack
        regions: eu-west-1
        target: XXXXXXXXXXXX
        properties:
          root_dir: ./second-stack
          param_overrides:
            - inputs: outputs
              param: FirstParam
              key_name: FirstParam
            - inputs: outputs
              param: SecondParam
              key_name: SecondParam
```

This will lead into the following CloudFormation snippet:

```json
{
  "Actions": [
    {
      "ActionTypeId": {
        "Category": "Deploy",
        "Owner": "AWS",
        "Provider": "CloudFormation",
        "Version": "1"
      },
      "Configuration": {
        "ActionMode": "CHANGE_SET_REPLACE",
        "StackName": "my-stack",
        "ChangeSetName": "my-stack",
        "TemplateConfiguration": "my-stack-build::./second-stack/params/account_name_eu-west-1.json",
        "Capabilities": "CAPABILITY_NAMED_IAM,CAPABILITY_AUTO_EXPAND",
        "RoleArn": "arn:aws:iam::XXXXXXXXXXXX:role/adf-cloudformation-deployment-role",
        "TemplatePath": "my-stack-build::./second-stack/template_eu-west-1.yml",
        "ParameterOverrides": "{\"FirstParam\": {\"Fn::GetParam\": [\"outputs\", \"outputs.json\", \"FirstParam\"]}, \"KeyArn\": {\"Fn::GetParam\": [\"outputs\", \"outputs.json\", \"SecondParam\"]}}"
      },
      "InputArtifacts": [
        {
          "Name": "my-stack-build"
        },
        {
          "Name": "outputs"
        },
        {
          "Name": "outputs"
        }
      ],
      "Name": "account_name-eu-west-1-create",
      "Region": "eu-west-1",
[...]

PR #446, opened 2022-03-08
fix: do not add duplicate inputs

Nr18:
*Issue #, if available:* #445

*Description of changes:*

When using the `param_overrides` functionality. And you use 2 parameters from the same source. The artifact is added twice to the action. This causes the pipeline to fail.


sbkok (changes requested):
Thanks for providing a fix for this issue.

Could you please add test cases to the following file [0] to verify that this works correctly?
That way we know it will continue to work in the future, even if this part gets refactored.

* [0] https://github.com/awslabs/aws-deployment-framework/blob/0723ddf4eaf55888ae780dc48873f0ec4766cfbd/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/tests/test_adf_codepipeline_input_artifacts.py

Nr18:
@sbkok I added 2 tests:
* 1 to make sure that inputs are not added twice (the original fix)
* 2 when using different inputs it should be added.

Also added some context on how to run the top build locally.

sbkok (approved):
LGTM, thanks for adding the tests too.

javydekoning (approved):
Manually ran our latest CI on this. All checks are passing.

PR LGTM! Thanks for contributing.

Issue #449, opened 2022-03-09
Allow override of the TemplateConfiguration for the CloudFormation provider


Nr18:
When using the ServiceCatalog as a deployment provider you vane the ability to supply a `configuration_file_path`. When you use the `generate_params.py` functionality with a single template in the pipeline you will not have any issues. 

But when you have multiple templates that you would like to deploy as part as a single deployment and the CloudFormation Paramaters of those stacks are not equal you will run into issues.

By supporting the `configuration_file_path` for a CloudFormation deployment target you would make it possible to specify specific configurations for these "other" stacks.

A use case for this is combining a delegation role in the master account and the implementation of the service in another account. By keeping them in a single pipeline and repo you make it easy to maintain and the execution order is guaranteed.

sbkok:
If I get it right, you are using the CloudFormation provider, and you want to have multiple stacks within the same repository.

Did you consider using the Mono Repo structure, as explained here?
https://github.com/awslabs/aws-deployment-framework/tree/0723ddf4eaf55888ae780dc48873f0ec4766cfbd/samples/sample-mono-repo

The `root_dir` property allows you to specify which stack you are deploying. Storing each stack in a separate directory.
This way can use one pipeline per stack, or combine all in a single stack.

If you combine all in a single stack, you would change directories into each stack, run the `generate_params.py` script and go to the next directory. For the deployment targets, you can set the `root_dir` per target.

Would this resolve your issue?

Nr18:
Hi @sbkok,

Yes I did had a look at the monorepo example, but that requires 2 pipelines and then you cannot pass the output of the first CloudFormation stack to the next.

I tried setting the root_dir and executing the `generate_params.py` but this failed. I will reproduce it and post the details, if I remember correctly it had import issues when executed. So I figured it was not a supported scenario. But from your reaction I assume it should work? In that case I might have found a bug then.

I will come back to this!

Nr18:
```txt
[Container] 2022/03/11 11:55:16 Running command cd ./subscription
[Container] 2022/03/11 11:55:16 Running command python ../adf-build/generate_params.py
Traceback (most recent call last):
File "../adf-build/generate_params.py", line 17, in <module>
from resolver import Resolver
File "/codebuild/output/src140950311/src/adf-build/resolver.py", line 11, in <module>
from s3 import S3
ModuleNotFoundError: No module named 's3'
 
[...]

PR #450, opened 2022-03-09
feat: allow `TemplateConfiguration` override from the target properties

Nr18:
*Issue #, if available:* #449

*Description of changes:*
By specifying a `configuration_file_path` property on the target properties you control the configuration used. This could be useful when you do not want to use the `generate_params.py` script. Or if you have multiple stacks with different parameters that you want to deploy from a single pipeline.


StewartW (approved):

sbkok:
I was going through the conflicts to update this PR.
However, it seems that I implemented this feature too. Back when I refactored the generate params script.
Apologies for not aligning with this PR back then. The current parameter name is called: `params_filename`.

On the bright side, you can use this feature with the current v3.2.0 right away.
No need to wait for the next release.

PR #459, opened 2022-04-02
Fix #458

### LINKED ISSUE #458: bug: AccountFileProcessingFunction failing due to Unknown keyword argumen
Logs:

```
[ERROR] ValueError: Unknown keyword argument: object_key
Traceback (most recent call last):
  File "/var/task/process_account_files.py", line 109, in lambda_handler
    account_file = get_file_from_s3(get_details_from_event(event), s3_resource)
  File "/var/task/process_account_files.py", line 43, in get_file_from_s3
    s3_object = s3_resource.Object(**s3_object)
  File "/var/runtime/boto3/resources/factory.py", line 473, in create_resource
    return partial(resource_cls, *positional_args,
  File "/var/runtime/boto3/resources/base.py", line 111, in __init__
    raise ValueError('Unknown keyword argument: {0}'.format(name))
```

`s3_resource.Object(**s3_object)` takes `bucket_name` and `key` as arguments as per docs: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html

Seems to be implemented incorrectly: https://github.com/awslabs/aws-deployment-framework/blob/e83c2f9cd7fe75e92abf59ec9a8e213acf8b0330/src/lambda_codebase/account_processing/process_account_files.py#L34-L37 

`object_key` should be `key`
 
It also tries to read from a stream, but binary mode doesn't accept encoding. 

```
[ERROR] ValueError: binary mode doesn't take an encoding argument
Traceback (most recent call last):
  File "/var/task/process_account_files.py", line 110, in lambda_handler
    account_file = get_file_from_s3(get_details_from_event(event), s3_resource)
  File "/var/task/process_account_files.py", line 45, in get_file_from_s3
    with tempfile.TemporaryFile(encoding='utf-8') as file_pointer:
  File "/var/lang/lib/python3.9/tempfile.py", line 610, in TemporaryFile
    return _io.open(fd, mode, buffering=buffering,
```

I'd suggest using:

```python
        with tempfile.TemporaryFile(mode='w+b') as file_pointer:
            s3_object.download_fileobj(file_pointer)

            # Move pointer to the start of the file
            file_pointer.seek(0)

            return yaml.safe_load(file_pointer)
```

Finally, the function seems to be invoked on all files. Even non yml/yaml files (e.g. the readme.md). When that happens it obviously fails parsing the YAML.

javydekoning:
*Issue #, if available:*
#458 

*Description of changes:*
- Fixed keyword error `key` instead of `object_key`.
- Remove encoding for `TemporaryFile`, which is not supported when streaming binary.
- Added S3Key filter to only invoke the Lambda on `.yml` files (no others are supported). 
- Added debug logging. 

### DIFF
--- a/src/account_processing.yml
+++ b/src/account_processing.yml
@@ -177,6 +177,17 @@ Resources:
                 Rules:
                 - Name: suffix
                   Value: yml
+        S3Event2:
+          Type: S3
+          Properties:
+            Bucket:
+              Ref: ADFAccountBucket
+            Events: s3:ObjectCreated:*
+            Filter:
+              S3Key:
+                Rules:
+                - Name: suffix
+                  Value: yaml
   AccountAliasConfigFunctionRole:
     Type: "AWS::IAM::Role"
     Properties:
--- a/src/account_processing.yml
+++ b/src/account_processing.yml
@@ -172,7 +172,11 @@ Resources:
             Bucket:
               Ref: ADFAccountBucket
             Events: s3:ObjectCreated:*
-
+            Filter:
+              S3Key:
+                Rules:
+                - Name: suffix
+                  Value: yml
   AccountAliasConfigFunctionRole:
     Type: "AWS::IAM::Role"
     Properties:
--- a/src/account_processing.yml
+++ b/src/account_processing.yml
@@ -166,7 +166,7 @@ Resources:
       FunctionName: AccountFileProcessorFunction
       Role: !GetAtt AccountProcessingLambdaRole.Arn
       Events:
-        S3Event:
+        S3YmlSuffixEvent:
           Type: S3
           Properties:
             Bucket:
@@ -177,7 +177,7 @@ Resources:
                 Rules:
                 - Name: suffix
                   Value: yml
-        S3Event2:
+        S3YamlSuffixEvent:
           Type: S3
           Properties:
             Bucket:
--- a/src/lambda_codebase/account_processing/process_account_files.py
+++ b/src/lambda_codebase/account_processing/process_account_files.py
@@ -10,8 +10,8 @@ invokes the account processing step function per account.
 import json
 import os
 import tempfile
-import yaml
 import logging
+import yaml
 
 from yaml.error import YAMLError
 
@@ -103,7 +103,7 @@ def start_executions(sfn_client, processed_account_list):
 
 def lambda_handler(event, _):
     """Main Lambda Entry point"""
-    LOGGER.debug(json.dumps(s3_object, indent=2) if LOGGER.isEnabledFor(logging.DEBUG) else "--data-hidden--")
+    LOGGER.debug(json.dumps(event, indent=2) if LOGGER.isEnabledFor(logging.DEBUG) else "--data-hidden--")
     sfn_client = boto3.client("stepfunctions")
     s3_resource = boto3.resource("s3")
 
--- a/src/lambda_codebase/account_processing/process_account_files.py
+++ b/src/lambda_codebase/account_processing/process_account_files.py
@@ -11,6 +11,8 @@ import json
 import os
 import tempfile
 import yaml
+import logging
+
 from yaml.error import YAMLError
 
 import boto3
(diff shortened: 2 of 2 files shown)


StewartW (dismissed):

src/lambda_codebase/account_processing/process_account_files.py
```diff
@@ -100,6 +101,7 @@ def start_executions(sfn_client, processed_account_list):
 
 def lambda_handler(event, _):
     """Main Lambda Entry point"""
+    LOGGER.debug(json.dumps(event, indent=2))
```
sbkok, line 104: Lets only do a `json.dumps` when debug mode is enabled.

src/account_processing.yml
```diff
@@ -172,7 +172,11 @@ Resources:
             Bucket:
               Ref: ADFAccountBucket
             Events: s3:ObjectCreated:*
-
+            Filter:
+              S3Key:
+                Rules:
+                - Name: suffix
+                  Value: yml
```
sbkok, line 179: We need to match `yaml` as well I think.

src/lambda_codebase/account_processing/process_account_files.py
```diff
@@ -33,14 +33,15 @@ def get_details_from_event(event: dict):
...
     return {
         "bucket_name": bucket_name,
-        "object_key": object_key,
+        "key": object_key,
     }
 
 
 def get_file_from_s3(s3_object: dict, s3_resource: boto3.resource):
     try:
+        LOGGER.debug(json.dumps(s3_object))
```
sbkok, line 42: Same comment as below, json.dumps only on debug.

sbkok (changes requested):

src/account_processing.yml
javydekoning: Agreed.

As per: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfiguration-config-filter-s3key.html

> The same type of filter rule cannot be used more than once. For example, you cannot specify two prefix rules.

This likely means I need to add another obj in the Events list as per: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-s3.html#sam-function-s3-events

src/account_processing.yml
```diff
@@ -172,7 +172,22 @@ Resources:
...
+        S3Event2:
```
StewartW, line 180: Can these have better names that just S3Event and S3Event2?

javydekoning: What about S3YamlSuffixEvents and S3YmlSuffixEvents?

StewartW: Works for me!

javydekoning: Done

StewartW (approved):

src/lambda_codebase/account_processing/process_account_files.py
```diff
@@ -33,14 +35,15 @@ def get_details_from_event(event: dict):
...
     return {
         "bucket_name": bucket_name,
-        "object_key": object_key,
+        "key": object_key,
     }
 
 
 def get_file_from_s3(s3_object: dict, s3_resource: boto3.resource):
     try:
+        LOGGER.debug(json.dumps(s3_object, indent=2) if LOGGER.isEnabledFor(logging.DEBUG) else "--data-hidden--")
```
sbkok, line 44: Could you add some context on where this debug is located, like so?

```suggestion
        LOGGER.debug(
            "Reading YAML from S3: %s",
            json.dumps(s3_object, indent=2) if LOGGER.isEnabledFor(logging.DEBUG) else "--data-hidden--"
        )
```

src/lambda_codebase/account_processing/process_account_files.py
```diff
@@ -100,6 +103,7 @@ def start_executions(sfn_client, processed_account_list):
 
 def lambda_handler(event, _):
     """Main Lambda Entry point"""
+    LOGGER.debug(json.dumps(event, indent=2) if LOGGER.isEnabledFor(logging.DEBUG) else "--data-hidden--")
```
sbkok, line 106: Do we still need this debug?
If so, could you add some context here too?

sbkok (changes requested):

sbkok (approved):
LGTM, Thank you!!

StewartW (approved):

Issue #468, opened 2022-04-08
feat: set codecommit repository description


Nr18:
We want to set the CodeCommit description from the deployment map. And looking at the ADF codebase it is partially implemented already. So I figured that I would complete the feature ;-) since we want to make use of it.

The implementation is really simple, see: #469

But I noticed that the `codecommit:UpdateRepository` action is part of the IAM permissions for the `adf-automation-role`. But this permission does not exist at all. You have `codecommit:UpdateRepositoryName` and `codecommit:UpdateRepositoryDescription`.

Changing a repository name is tricky, so that should not be there (I think) but the `codecommit:UpdateRepository` action should be changed to `codecommit:UpdateRepositoryDescription` IMHO.

sbkok:
Thank you for your patience. I am happy to inform you that this feature has been [released as part of v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues related to this feature.

PR #469, opened 2022-04-08
feat: use default description as CodeCommit repository description

Nr18:
*Issue #, if available:* #468

*Description of changes:*
When a pipeline description is given it was already propagated to the CodeCommit repository. Only the schema did not allow to provide the description, and the IAM policy contained an invalid action and did not allow description updates.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/repo.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/repo.py
@@ -81,6 +81,6 @@ class Repo:
         if _repo_exists and not _stack_exists:
             # return when the repository exists without a stack (previously made)
             return
-        else:
-            LOGGER.info(f"Ensuring State for CodeCommit Repository Stack {self.name} on Account {self.account_id}")
-            cloudformation.create_stack()
+
+        LOGGER.info(f"Ensuring State for CodeCommit Repository Stack {self.name} on Account {self.account_id}")
+        cloudformation.create_stack()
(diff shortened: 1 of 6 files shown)


StewartW (approved):

javydekoning (approved):
LGTM!

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/repo.py
```diff
@@ -75,16 +75,12 @@ def create_update(self):
...
         if _repo_exists and not _stack_exists:
+            # return when the repository exists without a stack (previously made)
             return
-        if not _repo_exists and not _stack_exists:
-            LOGGER.info(
-                'Ensuring State for CodeCommit Repository Stack %s on Account %s',
-                self.name,
-                self.account_id,
-            )
-            cloudformation.create_stack()
```
sbkok, line 90: The change here will perform updates to the CodeCommit CloudFormation stack.
Before we only created the repository, we didn't update it. I feel we should protect the code from being deleted by accident.
So we will also need an update policy to retain the repository just in case for that to be safe.
Can you add that?

See https://github.com/awslabs/aws-deployment-framework/blob/9420cdb908287dc545eff5a5bb3018f51b091046/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/templates/codecommit.yml

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/repo.py
Nr18: Sure added it in [f70bb46](https://github.com/awslabs/aws-deployment-framework/pull/469/commits/f70bb46d1719d21e1b7a0766d4b082948f25590b)

StewartW (approved):
lgtm

sbkok (approved):
LGTM, thank you!

PR #484, opened 2022-08-12
Fix: Wave deployment input data

sbkok:
**Why?**

When generating the waves, it does so in the input generation.
When it was shared to the actual deployment of the CodePipeline stack, the data types did not match.

**What?**

This pull request adds comments and restructures the wave generation at the input side to make it work.

This PR also fixes a typo and optimizes the generation of waves by using `yield` only.

Co-authored by: StewartW

---


sbkok:
Changes in this PR are moved towards the pipeline state machine PR #424.
I will remove the line changes from this PR that have been moved and add only remaining files if any or close the PR if none are left.

sbkok:
This Fix makes the current `master` branch work with wave deployments.
However, considering we are changing the pipeline generation in PR #424, @StewartW was so kind to incorporate the fixes in that PR directly instead. Hence I will close this PR once #424 is merged.

PR #485, opened 2022-08-12
Fix: Lookup correct Organization API Region

sbkok:
**Why?**

At the moment, the organization API calls are made to the region where the `organizations.py` code is running.

However, which region to use depends on the partition that is in use.
For the US gov-cloud regions, it should use the `us-gov-west-1` region.
While the `aws` partition regions should use `us-east-1`.

**What?**

The code will determine the partition and return the correct region to use for the Organizations API.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
@@ -15,9 +15,10 @@ from botocore.exceptions import ClientError
 from errors import RootOUIDError
 from logger import configure_logger
 from paginator import paginator
+from partition import get_organization_api_region
 
 LOGGER = configure_logger(__name__)
-REGION_DEFAULT = os.getenv('AWS_REGION')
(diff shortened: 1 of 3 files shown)


StewartW (approved):

javydekoning (changes requested):
What about cn? Should we not also handle aws-cn (cn-northwest-1) as per https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/organizations.html

Or otherwise just throw an error instead of defaulting to us-east-1?

sbkok:
Unfortunately, last time I checked the `cn` partition does not support all the services required for ADF to be installed.
I don't want to give a false hope to those that read these files and expect the `cn` partition to work out of the box.

javydekoning:
@sbkok , ack! Then what do you think about adding an if statement for the `aws` partition and Throwing an error on any other unsupported AWS region?

sbkok:
Good idea, I will add that.

javydekoning (approved):
Thanks for the changes and including tests! LGTM!

StewartW (approved):

Issue #488, opened 2022-08-17
Bug in CodeBuild 'size' and 'privileged' properties when using to deploy


joaorodrig:
I found a bug which is quite easy to correct. Go to this file:

https://github.com/awslabs/aws-deployment-framework/blob/master/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py

In lines 46 and 50, the pipeline attributes are being retrieved from the build part, as opposed to coming from the deploy part.

As a temporary workaround, the following configuration allows to pass the "size" and "privilege" settings to the deployment, even if a build is not needed.

```
  - name: my_pipeline
    default_providers:
      source:
        provider: codecommit
        properties:
          account_id: "MY_ACCOUNT_ID"
          repository: "my_repository"
          branch: master
      build:
        enabled: False
        properties:
          size: medium # this can be used to pass the property to the deploy phase, because of the error described above
      deploy:
        provider: codebuild
        properties:
          image: "STANDARD_5_0"
          timeout: 60
```

sbkok:
Thank you @joaorodrig, for reporting this issue and for providing your analysis into how to fix it quickly!
I've opened a PR to get this fixed in our next release.

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.0](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.0) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues with the latest release.

PR #489, opened 2022-08-17
Fix incorrect CodeBuild build/deploy references in params

sbkok:
Issue: #488.

**What?**

Updated references and made sure line length would not exceed the maximum on
the lines that were changed.

---


StewartW (dismissed):

sbkok:
I found another issue while looked at the code once more.
I've added the following fix:

Fix CodeBuild privilege override with target parameters

**Why?**

At the moment, if you would set privilege to `True` on the default deploy provider properties and try to override those to `False` in a specific target, it will still have privilege enabled.

**What?**

Updated the logic to only override if it was not specified, not when it is `False`.

StewartW (approved):

PR #490, opened 2022-08-17
Refactor line length and minor code style changes throughout ADF

sbkok:
**Why?**

The files touch upon in this change set did not follow the correct line
length that is defined by the EditorConfig.

While going through the lines, I also noticed a few comments and code blocks
that needed to be improved.

**What?**

* Changed the line lengths and minor adjustments in comments where required.
* Added new lines before resource definitions in the CloudFormation templates so it is easier to navigate and read.
* Minor refactoring of the code where needed.

---


javydekoning (approved):
LGTM, we should consider adding a linter for this later.

sbkok:
Absolutely, I am planning to introduce `black` after we released 3.2.0. So it does not conflict with our PRs that introduce large change sets at the moment.

StewartW (approved):

PR #495, opened 2022-08-22
fixing linter issues introduced by rdk examples

javydekoning:
*Issue #, if available:*
N/A

*Description of changes:*
This PR makes style adjustments that do not satisfy our linters. Unfortunately the `rdk` samples PR merge crossed the merge of the linter, hence they where not caught. Specifically this rule:

https://github.com/awslabs/aws-deployment-framework/blob/master/.editorconfig#L30-L31

Also added: 

```yaml
CLOUDFORMATION_CFN_LINT_FILE_EXTENSIONS: [".yml", ".yaml"]
```

To the linter config to bring it inline with our `cfn-lint` configuration. This repo doesn't have `json` CloudFormation templates, but the `rdk` samples have partial templates in `json` that are otherwise incorrectly flagged.
 
### DIFF
--- a/.mega-linter.yml
+++ b/.mega-linter.yml
@@ -30,3 +30,5 @@ FILEIO_REPORTER: false
 JSON_PRETTIER_PRE_COMMANDS:
   - command: npm install prettier-plugin-multiline-arrays@1.1.0
     cwd: "workspace"
+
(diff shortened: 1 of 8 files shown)


StewartW (approved):

sbkok (approved):
Thanks for fixing these!

PR #496, opened 2022-08-25
Update CDK and others + use Python 3.9 in CodeBuild

sbkok:
**Why?**

Currently, the CodeBuild steps that process the bootstrap and pipelines logic
are configured to use `STANDARD_6_0` with Python 3.9. While `STANDARD_6_0` only
supports Python 3.10.

Our test suite is running against Python 3.9.
And AWS Lambda does not support Python 3.10 just yet.
Therefore I suggest to Python 3.9 for all logic instead. Changing the image back
to `STANDARD_5_0` which does support this version of Python.

**What?**

Updated the required linters, CDK, and other libraries to the latest version and
configured all CodeBuild environments to use the same version (`STANDARD_5_0`).
Including our samples.

---


javydekoning (approved):
Thanks Simon. LG!

StewartW (approved):

PR #502, opened 2022-08-25
Fix correct use of partitions

sbkok:
**Why?**

Some of the pull requests might have overlapped with the pull request that introduced support for the `aws-us-gov` partition.

By searching for `:aws:`, I found a number of references that were still hard coding the `aws` partition.

**What?**

Adding support for dynamic partitions where it was not supported yet.

---


javydekoning (approved):
Thanks, looks good!

StewartW (approved):

PR #506, opened 2022-08-30
Fix syncing deployment_map files to s3

sbkok:
**Why?**

When it tried to sync a deployment_map.yml file that did not exist in the
pipelines repo, it would fail with exit code 1.

Additionally, the files directly inside the deployment_maps folder were not
synced, only nested folders were synced correctly.

**What?**

* Moved install commands to the install phase in the CodeBuild buildspec.
* Added a check to test whether the folder/file exists before syncing.
* Added echo messages to explain what and why it does or doesn't sync.
* Added or statement to ensure that it will succeed with exit 0 when
  files/folders are missing.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
@@ -762,15 +762,14 @@ Resources:
               runtime-versions:
                 python: 3.9
                 nodejs: 14
-            pre_build:
               commands:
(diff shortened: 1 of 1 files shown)


StewartW (approved):
Looks good. 

As a sidenote - should we look at moving these build commands out into a separate script file and store it in the SHARED_MODULES_BUCKET for future versions?

sbkok:
Sounds good to me. You mean the copy/sync steps right?
I can take a look at that.

StewartW:
> Sounds good to me. You mean the copy/sync steps right? I can take a look at that.

Yeah, it's not required for the PR though.

javydekoning (approved):
Much better this way!

PR #508, opened 2022-08-31
Add support to change main branch on bootstrap and pipelines repositories

sbkok:
## Why?

We set a goal to remove the word `master` from our repository as much as possible. The bootstrap repository initialized to the `master` branch by default and creates pull requests against the `master` branch too.

Additionally, the CodePipeline should monitor the correct branch to perform the changes required when the main branch gets updated.

## What?

For any new installation, the bootstrap repository should default to the `main` branch.

While any existing installation of ADF should allow the administrators to move to the `main` branch easily.
To enable this, when ADF is updated, it will perform a check which branch is configured as the default branch on the repository (`aws-deployment-framework-bootstrap` and `aws-deployment-framework-pipelines`). It will create the pull request against the default branch if it needs to be updated.

If you want to switch from one branch to another, you need to create a new branch from the current default branch.

For example, switching from `master` to `main`:
Simply running these commands from inside the `aws-deployment-framework-bootstrap` or `aws-deployment-framework-pipelines` repository that is cloned locally should do it:

```bash
git fetch
git checkout origin/master
git push HEAD:origin/main
```

Navigate to the CodeCommit repository and update the default branch of the repository to the new branch. Make sure to click the `Save` button underneath the default branch setting to save it.

---


sbkok:
PS another PR will be created shortly that will update the docs to update the language to be more inclusive.
Like updating usage of the `master` account to `management` account, etc.

StewartW (approved):
Looks good. 
Good use of custom resources.

javydekoning (approved):
Echoing what @StewartW said, nice improvement. My only concern is the amount of extra code to maintain to support this.

PR #510, opened 2022-08-31
Improvement: Only invoke pipeline deletion when needed

sbkok:
## Why?

At the moment, the InvokeDeleteStateMachine task is called even when no pipelines need to be deleted. This causes the actual state machine that deletes these tasks to fail. As it cannot iterate over an empty list.

## What?

Introducing a choice, that will only invoke the delete pipelines state machine when there is a need to do so.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
@@ -420,7 +420,18 @@ Resources:
                 "BackoffRate": 1.5,
                 "MaxAttempts": 10
               }],
-              "Next": "InvokeDeleteStateMachine"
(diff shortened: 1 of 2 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -420,7 +420,18 @@ Resources:
...
                 "MaxAttempts": 10
               }],
-              "Next": "InvokeDeleteStateMachine"
+              "Next": "AnyPipelineToBeDeleted?"
+            },
+            "AnyPipelineToBeDeleted?": {
+              "Type": "Choice",
+              "Choices": [
+                {
+                  "Variable": "$.pipelines_to_be_deleted",
```
StewartW, line 429: 🤌

StewartW (approved):

javydekoning (approved):

PR #513, opened 2022-08-31
Add retry logic to SF Lambda invocations

sbkok:
## Why?

When the Lambda service is experiencing issue, it could affect the execution of the Step Function. Most of the errors it would fail with could be recovered from.

Additionally, with large ADF installations the Lambda function could timeout when an update is performed. As it was not able to recover, the whole update would fail in a hard-to-fix state.

## What?

* Added retry logic around Lambda service issues.
* Plus added a retry mechanism to recover from Lambda function executions that ran for too long. To retry those automatically too.
* Improved log messages, so they output in the same structure within the bootstrap State Machine. The logs were using a mix of output styles, making it hard to debug.
* Fixed indentation of State Machine definitions in our CloudFormation stacks, some were using a mixture of 4 spaces and 2 spaces, while we should use 2 spaces here.

---


javydekoning (approved):

src/lambda_codebase/account_bootstrap.py
```diff
@@ -25,6 +25,7 @@
 REGION_DEFAULT = os.environ["AWS_REGION"]
 PARTITION = get_partition(REGION_DEFAULT)
 LOGGER = configure_logger(__name__)
+DEPLOY_TIME_IN_MS = 5 * 60 * 1000
```
StewartW, line 28: What is this magic number?

StewartW (approved):

PR #515, opened 2022-08-31
Fix updating account alias when needed

sbkok:
## Why?

At the moment, the account management state machine will attempt to add the specified alias to the account whether it is present or not.

In case the alias already exists, it will fail.

## What?

This change will ensure the aliases configured on the account are removed when they are no longer configured (you can only have one, but the API supports multiple). It will then add the alias if the alias that needs to be present is
missing.

Updated maximum number of attempts in the State Machines to max 10, as 30 would imply multiple days of retries.

---


javydekoning (approved):
Thanks Simon, makes more sense like this.

StewartW (approved):

PR #517, opened 2022-09-01
Add support for the CodeBuild provider to run inside a VPC

sbkok:
## Why?

To add support for build/deploy steps that use CodeBuild and need to connect to internal resources that are only available through a VPC.

This feature allows you to configure which VPC to connect to, using a specific list of subnets and security groups. The latter helps ensure that the connection is only allowed to connect when a specific security group is used.

## What?

Added three properties to the CodeBuild provider:
- vpc_id (string)
- subnet_ids (list of ids (strings))
- security_group_ids (list of ids (strings))

When you configure a pipeline to use a specific VPC, you need to specify the subnets to use as well.

In case you don't specify any security groups, they will be automatically created for you. These security groups allow all egress traffic by default. It is recommended to use a specific security group so you can strictly define what egress traffic is allowed.

__Please note__: When you configure a specific CodeBuild environment in a pipeline to use a VPC it cannot be removed easily. To remove the VPC from the CodeBuild environment, you need to remove the pipeline stack and instruct ADF to generate the pipelines again (release trigger on `aws-deployment-framework-pipelines` in CodePipeline).

Documentation has been added and a sample that shows how to setup a pipeline to use a specific VPC.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/generate_pipeline_stacks.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/generate_pipeline_stacks.py
@@ -26,7 +26,6 @@ ADF_LOG_LEVEL = os.environ["ADF_LOG_LEVEL"]
 def main():
     LOGGER.info("ADF Version %s", ADF_VERSION)
     LOGGER.info("ADF Log Level is %s", ADF_LOG_LEVEL)
-    _threads = []
     _templates = glob.glob("cdk_inputs/*.json")
     for template_path in _templates:
         with open(template_path, encoding="utf-8") as template:
--- a/docs/providers-guide.md
+++ b/docs/providers-guide.md
@@ -307,6 +307,106 @@ Provider type: `codebuild`.
   > Note: Either specify the `spec_inline` or the `spec_filename` in the
   > properties block. If both are supplied, the pipeline generator will throw
   > an error instead.
+- *vpc_id* *(String)* defaults to none.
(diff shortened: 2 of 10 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/generate_pipeline_stacks.py
```diff
@@ -26,7 +26,6 @@
 def main():
     LOGGER.info("ADF Version %s", ADF_VERSION)
     LOGGER.info("ADF Log Level is %s", ADF_LOG_LEVEL)
-    _threads = []
```
sbkok, line 29: This was not used anywhere, hence I removed it.

StewartW (approved):

javydekoning (commented):
Shouldn't `spec_filename` default to `buildspec.yml` if  both `spec_inline` and the `spec_filename` is not provided?

sbkok:
By default, when they are not supplied, they are not forwarded to CDK and thereby CloudFormation either.
Which defaults to `buildspec.yml`, see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-source.html#cfn-codebuild-project-source-buildspec

javydekoning (approved):
LGTM, I personally don't like the `Blockquotes` style (`>`) in the  unordered lists in the docs. It requires a prefix on every line and is harder to format (like line wrapping if we stick to linelength=80) and mix with codefences. 

I have changed this in my latest PR #521 so we might have to change that if we choose to merge #521

Issue #518, opened 2022-09-01
ADF-Bootstrap CodeBuild fails directly after creating a new account using adf-accounts logic


sbkok:
## Problem

CodeBuild fails directly after creating a new account.

### Error 1
> Failed to create the change set for adf-global-base-development | (cloudformation.py:282)

### Error 2
> Failed to update its base stack due to missing parameters (deployment_account_id or kms_arn), ensure this account has been bootstrapped correctly by being moved from the root into an Organizational Unit within AWS Organizations

### Error 3
> 2022-09-01 15:28:21,617 | DEBUG | s3 | Nothing could be found for regional.yml when traversing the bucket | (s3.py:201)
> 
> Traceback (most recent call last):
> File "/codebuild/output/src013183384/src/adf-build/shared/python/cloudformation.py", line 277, in _create_change_set
> self.client.create_change_set(**change_set_params)
> File "/root/.pyenv/versions/3.9.12/lib/python3.9/site-packages/botocore/client.py", line 508, in _api_call
> return self._make_api_call(operation_name, kwargs)
> File "/root/.pyenv/versions/3.9.12/lib/python3.9/site-packages/botocore/client.py", line 915, in _make_api_call
> raise error_class(parsed_response, operation_name)
> botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateChangeSet operation: Unable to fetch parameters [deployment_account_id] from parameter store for this account.
>
> The above exception was the direct cause of the following exception:
>
> Traceback (most recent call last):
> File "/codebuild/output/src013183384/src/adf-build/main.py", line 235, in worker_thread
> cloudformation.create_stack()
> File "/codebuild/output/src013183384/src/adf-build/shared/python/cloudformation.py", line 390, in create_stack
> create_change_set = self._create_change_set()
> File "/codebuild/output/src013183384/src/adf-build/shared/python/cloudformation.py", line 290, in _create_change_set
> raise GenericAccountConfigureError(error) from error
> errors.GenericAccountConfigureError: An error occurred (ValidationError) when calling the CreateChangeSet operation: Unable to fetch parameters [deployment_account_id] from parameter store for this account.
>
> The above exception was the direct cause of the following exception:
>
> Traceback (most recent call last):
> File "/codebuild/output/src013183384/src/adf-build/main.py", line 381, in <module>
> main()
> File "/codebuild/output/src013183384/src/adf-build/main.py", line 357, in main
> thread.join()
[...]

PR #530, opened 2022-09-14
Sync Step Function input files when content changed only with exec id metadata

sbkok:
> Step 1 of fixing #518.

## Why?

By using `aws s3 sync/cp`, it would copy the files when these were changed. However, as the file metadata is also taken into account, it would upload also if the content did not change.

Additionally, as described in #518, we would like to insert metadata when a file is changed. If we would rely on the `aws s3 sync/cp` logic, it will also update the metadata if the metadata itself is changed. Therefore, we cannot add the necessary execution id to the files upon an upload only.

## What?

The `sync_to_s3.py` script is added to support syncing the files to S3. This script will:
1. Upload a single file, or walk through a directory recursively.
2. Check each of the files it finds and determines the SHA-256 hash of these.
3. Parse the S3 bucket with an optional prefix, to determine which objects exist.
4. If a file is missing, it will upload the file.
5. If a file exists as an object already, it will check if the SHA-256 hashes match. If they do not, it will upload the new version.
6. If an object exists, but the file does not exist, it will optionally delete the object from the S3 bucket.

When it uploads a file to S3, it will add the metadata that is requested through the `--upload-with-metadata` argument. Additionally, it will add the `sha256_hash` metadata to determine if the content changed.

The deployment maps and account configuration processes rely on AWS Step Functions. When these are synced, the process is updated to rely on the `sync_to_s3.py` script. This way we can retrieve the `execution_id` and insert that in the invocation id of the Step Function State Machine.

---


javydekoning (approved):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
```diff
@@ -767,10 +767,13 @@ Resources:
...
                 - pip install -r adf-build/requirements.txt -q -t ./adf-build
+            pre_build:
+              commands:
+                - mkdir -p deployment_maps
             build:
               commands:
-                - bash -c "[[ -e deployment_map.yml ]] && echo 'Copying deployment_map.yml' && aws s3 cp deployment_map.yml s3://$ADF_PIPELINES_BUCKET/deployment_map.yml || echo 'No deployment_map.yml, skipping copy'"
-                - bash -c "[[ -e deployment_maps ]] && echo 'Syncing deployment_maps folder' && aws s3 sync deployment_maps s3://$ADF_PIPELINES_BUCKET/deployment_maps || echo 'No deployment_maps folder, skipping sync'"
+                - python adf-build/helpers/sync_to_s3.py --upload-with-metadata execution_id=${!CODEBUILD_BUILD_NUMBER} deployment_map.yml s3://$ADF_PIPELINES_BUCKET/deployment_map.yml
+                - python adf-build/helpers/sync_to_s3.py --extension .yml --upload-with-metadata execution_id=${!CODEBUILD_BUILD_NUMBER} --recursive deployment_maps s3://$ADF_PIPELINES_BUCKET/deployment_maps
```
StewartW, line 776: Is it okay that this will be .yml and not support .yaml file endings?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/sync_to_s3.py
```diff
@@ -0,0 +1,820 @@
...
+    file_extension = options.get('--extension') or ""
+    s3_url = options.get('DESTINATION_S3_URL')
+    recursive = options.get('--recursive', False)
+    delete = options.get('--delete', False)
+
+    # Convert metadata key and value lists into a dictionary
+    # metadata_key_values = dict(map(
+    #     lambda idx: (options['<key>'][idx], options['<value>'][idx]),
+    #     range(len(options['<key>'])),
+    # ))
```
StewartW, line 797: Is this needed

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/sync_to_s3.py
```diff
@@ -0,0 +1,820 @@
...
+    if options["--verbose"] > 0:
+        LOGGER.setLevel(logging.DEBUG)
+    if options["--verbose"] > 1:
+        # Also enable DEBUG mode for other libraries, like boto3
+        logging.basicConfig(level=logging.DEBUG)
+
+    LOGGER.debug("Input arguments: %s", options)
+
+    local_path = options.get('SOURCE_PATH')
+    # file_extension = options.get('--extension') or ""
```
StewartW, line 787: Commented out code here

src/template.yml
```diff
@@ -1287,7 +1287,8 @@ Resources:
                 - aws s3 sync ./adf-build/shared s3://$DEPLOYMENT_ACCOUNT_BUCKET/adf-build --quiet
                 # Base templates:
                 - aws s3 sync . s3://$S3_BUCKET --quiet --delete
-                - aws s3 sync ./adf-accounts s3://$ACCOUNT_BUCKET --quiet
+                # Upload account files to the ACCOUNT_BUCKET
+                - python adf-build/shared/helpers/sync_to_s3.py --extension .yml --upload-with-metadata execution_id=${CODEBUILD_BUILD_NUMBER} --recursive adf-accounts s3://$ACCOUNT_BUCKET
```
StewartW, line 1291: I don't think this is the right env variable here.

StewartW: I think we want to add the execution ID from the CodePipeline to it via env variables and then use that

StewartW (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/sync_to_s3.py
sbkok: Good one, I replaced that. Will remove it.

src/template.yml
StewartW: Although you can query the execution Id using the pipeline state query. Just depends if we want those additional api calls and IAM permissions

sbkok: I looked into this, it is possible to get the CodePipeline execution id.

Unfortunately [CodeBuild environment variables](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html) do not expose the CodePipeline execution id as an environment variable just yet.

Apparently you can use the [`#{codepipeline.PipelineExecutionId}` variable in the CodePipeline definition](https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-variables.html#reference-variables-workflow).
We would need to update the CodePipeline definitions for this to be supported.
I would need to test how that would merge with setting environment variables on the CodeBuild project itself.

I am wondering though if this will update if you retry a single stage. Lets say CodeBuild fails due to another failure, will it get a new ID if you only retry that stage?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
sbkok: I've added support to add multiple extensions, so we can match with `.yml` and `.yaml`.

src/template.yml
StewartW: It looks like adding the variable from the pipeline won't impact project level variables: 
Screenshot 2022-09-16 at 09 17 48
See the lowercase codepipe_execution_id variable. 

I've tested single stage retry and the execution Id is the same, which is the desired outcome?

sbkok: Thanks for testing that. The number is a bit long though, but I guess we can use that.

I guess rerunning with the same ID would not hurt.
It would only use/apply the ID when there are content changes any way.

I will add the execution id to this PR, thanks!

sbkok: Done.

sbkok:
@StewartW I've updated the PR and included the ADF version to the metadata of the files as we discussed on Friday.
This allows us to trigger a new run of the Step Function State Machines when ADF is updated.

StewartW (approved):

javydekoning (approved):

PR #553, opened 2022-12-06
Update CDK to v1.188.1 and others

sbkok:
## Why?

For our next release that is closing in, we would like to make sure we rely on the latest CDK and other dependencies that are compatible.

## What?

Updated the required linters, CDK, and other libraries to the latest version that are compatible.

The CloudFormation linter complained about requiring a package step, which is not an issue in our pipelines at all. Added W3002 to the ignored checks for this reason.

---


javydekoning (approved):
Thanks!

StewartW (approved):

PR #554, opened 2022-12-06
Fix parameter validation upon install/update

sbkok:
## Why?

When installing or updating ADF, it did not validate the input given by the user. In some situations, this implied that it would attempt to install/update failing minutes later. Or it might not work at all after the CloudFormation deployment of the SAR template reports back as if all went well.

One of the parameters, the deployment region was particularly dangerous to leave blank on an update. As that would change the bucket name that would be used by ADF to share resources.

## What?

Added validations to the template.

The account alias regex was found [here](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateAccountAlias.html).
While the role name requirements were found [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-rolename).

---


StewartW (dismissed):

src/template.yml
```diff
@@ -30,48 +30,69 @@ Parameters:
...
-    Description: "The email address for the new or existing deployment account"
+    Description: >-
+      The email address for the new or existing deployment account. (Only required when
+      installing ADF for the first time.)
 
   DeploymentAccountAlias:
     Type: String
     Default: ""
-    Description: "(Optional) Example -> companyname-deployment (must be globally unique)"
+    AllowedPattern: "\\A([a-z0-9]([a-z0-9]|-(?!-))*[a-z0-9])?\\Z"
```
javydekoning, line 59: [Documentation](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateAccountAlias.html) uses a different Regex (see capture group start `(`)

`[a-z0-9](([a-z0-9]|-(?!-))*[a-z0-9])?`

Or is the capture group different because we allow an empty string to be passed?

src/template.yml
```diff
@@ -30,48 +30,69 @@ Parameters:
...
-      Example -> 123456789012  (Only required if you have an existing AWS Account
-      that you wish to use as the deployment account.)
+      Example -> 123456789012  (Only supported when installing ADF for the first time
+      and you have an existing AWS Account that you wish to use as the deployment
+      account. Leave blank otherwise.)
 
   DeploymentAccountMainRegion:
     Type: String
-    Default: ""
+    AllowedPattern: "\\A[a-z][a-z0-9\\-]*\\Z"
```
javydekoning, line 76: This would also match `uswest22322fasddfds`

Consider: `(us(-gov)?|ap|ca|cn|eu|sa)-(central|(north|south)?(east|west)?)-\d`

https://www.regextester.com/109163

src/template.yml
```diff
@@ -30,48 +30,69 @@ Parameters:
...
+    MinLength: 0
+    MaxLength: 63
+    Description: >-
+      Example -> companyname-deployment (Must be globally unique. Only required when
+      installing ADF for the first time.)
 
   DeploymentAccountId:
     Type: String
     Default: ""
     Description: >-
```
javydekoning, line 69: If we do pattern validation for other strings, let's consider doing it here as well.

`\d{12}`

src/template.yml
javydekoning: p.s. I don't think you need `\\A` and `\\Z`, Cloud-formation doesn't do partial matching.

The pattern must match the entire parameter value provided.

sbkok: Thanks for that regex, I will update it.
I will remove cn from the region prefix options, as we don't support that yet. Otherwise it looks like we support it.

src/template.yml
javydekoning: Well this is minor, but by moving the capture group you no longer allow single char aliases (that are allowed in the original regex, see first part `[a-z0-9]`). 

Instead you could just add another like this:

`([a-z0-9](([a-z0-9]|-(?!-))*[a-z0-9])?)?`

sbkok: I would be surprised if those aliases are still available. But happy to change it 😄

src/template.yml
```diff
@@ -42,6 +42,9 @@ Parameters:
   DeploymentAccountName:
     Type: String
     Default: ""
+    AllowedPattern: "[\\s\\S]"
```
javydekoning, line 45: `[\s\S]*`

Missing `*`

javydekoning: as per: https://docs.aws.amazon.com/organizations/latest/APIReference/API_Account.html

sbkok: Oh my, I really need coffee ;)
I referenced the same article in the commit, haha.

javydekoning (approved):
Looks good now, thank you!

PR #558, opened 2022-12-07
Fix syncing to S3 without a prefix

sbkok:
## Why?

When the `sync_to_s3.py` script is asked to sync files without a prefix set, it will instruct the S3 list-objects-v2 API to list all objects with the prefix `"/"`. This does not return anything. Hence the script thinks it needs to upload the files as they are missing.

## What?

* Added logic to only add the prefix if set.
* Added test to validate that the lookup without a prefix works as instructed.

---


javydekoning (approved):
Thanks!

StewartW (approved):

PR #559, opened 2022-12-07
Fix generate_params.py to rely on pipeline input definitions instead

sbkok:
## Why?

The `generate_params.py` script relied on the `account_ous.json` file that was uploaded to the shared modules bucket.

While ADF got support for newer account target selection mechanisms, the `account_ous.json` file structure was not updated accordingly.

To ensure future support and ease to maintain new account target selections it was decided to drop the `account_ous.json` and move to the pipeline input definition instead.

## What?

* Updated generate_params.py to rely on the pipeline input definition instead.
* Updated generate_params.py to ensure that parameters would be generated correctly. Refactored it, as before it was using a complex structure inside the resolver class. With the new changes, it relies on generating a new copy of the parameters with each step instead. Making it easier to test and harder to break.
* Refactored the resolver class to a version that is easier to understand and easy to extend using resolver classes.
* Added multiple test cases to ensure correctness of the generate_params.py script.
* The shared modules bucket was used to share the accounts and ous from the pipeline definition to the pipeline itself before. With this change, it is using the pipeline definition bucket and both the pipeline definition bucket and the shared modules bucket are now read only for the pipeline itself.
* There was another bucket accessible to the pipeline, but that was not used, hence dropping access to it.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
@@ -287,10 +287,17 @@ Resources:
             Resource:
               - !Sub arn:${AWS::Partition}:s3:::${PipelineBucket}
               - !Sub arn:${AWS::Partition}:s3:::${PipelineBucket}/*
+          - Effect: Allow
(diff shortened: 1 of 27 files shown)


StewartW (approved):
Looks great!

PR #562, opened 2022-12-09
Add clean-up of stale deployment map files in the pipeline bucket

sbkok:
## Why?

The identify-out-of-date-pipelines process will iterate over the S3 files and delete the stale pipelines that are no longer defined.

However, if you delete a deployment map file, the old deployment map would still exist in the pipelines bucket. Which will put the identify-out-of-date-pipeline process in the wrong direction, as it will think that these pipelines are still active.

## What?

Instructing the sync to S3 process to delete stale files in the S3 bucket, so deleted deployment map files are also deleted in the pipeline bucket.

---


javydekoning (dismissed):

sbkok:
Also added `DeleteObject` permissions to the CodeBuild Pipeline provisioner,
such that it can perform the required S3 calls.

javydekoning (approved):

StewartW (approved):

PR #563, opened 2022-12-09
Add support for CodeStar CodeBuild clone ref and update CodeStar setup

sbkok:
## Why?

The CodeStar provider did not support the `CODEBUILD_CLONE_REF` artifact mode yet. This is useful if you would like to access the git repository, for example to analyze the history or checkout a different branch.

For CodeStar Connections, it was trying to resolve the CodeStar Connection Arn from SSM Parameter Store at the pipeline stack generation. However, resolving parameters should be performed in the input generation step.

## What?

* Added support to define the `output_artifact_format` in the CodeStar source provider. Allowing `CODEBUILD_CLONE_REF` to be defined as the output format.
* Moved resolving the `codestar_connection_path` in SSM Parameter Store to the pipeline input generation. As this is where it should be performed. The Lambda function resolving these parameters has the required permissions already.
* Added support for our CloudFormation role to perform `codestar-connections:PassConnection`, such that CodeStar connections can be configured in the CodePipeline stack that it generates.

---

### DIFF
--- a/docs/providers-guide.md
+++ b/docs/providers-guide.md
@@ -211,6 +211,11 @@ Provider type: `codestar`.
     web hook as part of the pipeline. Read the CodeStar Connections
     documentation for more
     [information](https://docs.aws.amazon.com/dtconsole/latest/userguide/connections.html).
+- *output_artifact_format* - *(String)* default: `CODE_ZIP`
+  - The output artifact format. Values can be either `CODEBUILD_CLONE_REF` or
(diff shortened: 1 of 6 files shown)


javydekoning (approved):

StewartW:
Is this a regression, I feel like this was already added?

StewartW (approved):

sbkok:
It was added for CodeCommit, not for CodeStar apparently. This surprised me as well.

PR #567, opened 2022-12-14
Fix pipeline management logical id typo

sbkok:
## Why?

There is a typo in the pipeline management resource names. This PR fixes that.
Instead of `Mangement` it should be `Management`.

---


javydekoning (approved):

StewartW (approved):
Doh, thanks

PR #538, opened 2022-09-28
Feat/nested ou targets

StewartW:
# Why?

Describe why you are proposing these changes

*Issue #, if available:*

## What?

Description of changes:

Introduced the ability to resolve nested OUs within a path. 
eg: /**/* would resolve to all accounts with in the root. 

Allows for reduced overhead when defining targets and works with the excluded path options as well. 
---

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

PR #577, opened 2022-12-16
Implementing resolve_ou_recursively for deployment maps; adding '/Roo…

avolip:
PLEASE NOTE THIS PR TARGETS tags/v3.1.2

Adding two features to deployment maps targets section:

- Deploy into Root OU by using '/Root' as target path;
- New parameter 'resolve_ou_recursively' (bool) to recursively deploy into OUs

Reason:
- Many customers use ADF to deploy security baseline CloudFormation templates. In this way, customers can be sure their baseline is always applied to all accounts, regardless of tags.

### DIFF
--- a/docs/user-guide.md
+++ b/docs/user-guide.md
@@ -1,22 +1,30 @@
 # User Guide
 
-- [Deployment Map](#deployment-map)
-  - [Providers](#providers)
-  - [Targets Syntax](#targets-syntax)
-  - [Params](#params)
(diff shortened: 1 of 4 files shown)


StewartW:
Hey @avolip. Thanks for contributing!  Asides from the naming convention. Does this have any additional functionality to #538

avolip:
My bad I didn't notice there was already a PR on that! Shame on me for not checking :)

PR #581, opened 2022-12-22
Reduce number of cross-account access IAM API calls + add exp backoff retries

sbkok:
## Why?

1. In the EnableCrossAccountAccess process, it would call PutRolePolicy even when the policy was not modified.
Additionally, it would assume into the target account using STS for each region, while only making API calls to the global IAM service.

2. At the moment, with each region, it would perform the PutRolePolicy and GetRolePolicy calls required to update the target and deployment accounts.
On top of that, it would update the deployment account roles with every account it updated too.

3. The role used to manage the cross account access role was modifying itself.
Instead, it should use another role to manage the CloudFormation roles in the target accounts.

4. Ran into wrong wording on the Slack message, should not use Deny, but Reject a deployment. As the CodePipeline button states Reject.

### 1. Change EnableCrossAccountAccess process
* The IAM Class comments stated it was used across ADF, this is not the case.
* The implementation of the IAM class was very specific to enabling cross account access. Considering it is not used anywhere else, it is moved to the Lambda function code base in the deployment account.
* Renamed the IAM file and class to reflect what it is managing.
* Updated the implementation to rely on OOP best practices, creating different instances when we try to update another instead of overwriting the parameters using a singleton approach.
* Added tests to validate that the right permissions are set.
* Updated the retry logic in the Step Function state machine to retry on ClientErrors, with a higher wait period and exponential back off.

### 2. Reduce API calls to once per account
* Reducing the number of API calls by applying the deployment account role policies only when we update the deployment account.
* Additionally, we update the target accounts after iterating over the regions and determining the full list of S3 buckets and KMS keys to grant access to.

### 3. Role used to modify
* Added an update cross account role, that is assumed from the enable cross account access logic to update the CloudFormation roles.
* Removed adf-lambda-role, as that was only used by Send Slack Notifications, Check Pipeline Status, and the Enable Cross Access Role.
  * Restricted policies to bare minimum for each of these instead.
* Added comments explaining the subtle differences between target and deployment account cross account role management.

### 4. Slack notification class
* Fixed the wording and improved line-length issues.

---


StewartW (approved):

StewartW:
Looks great. Love it when we optimise these sort of api calls

javydekoning (approved):

PR #588, opened 2023-01-12
Fix bootstrap mirror pipelines

sbkok:
## Why?

An error was generated when the `aws-deployment-framework-bootstrap` is mirrored from another location, via a pipeline that is deployed with ADF. For example, hosted in Github, where the pipeline connects to Github via a CodeStar Connection and pushes the code to the local CodeCommit repository.

In this case, the `generate_params.py` file will look for the new `pipeline_input` attribute in the pipeline definition. While the pipeline definition was generated with a previous version of ADF, thereby using the `input` attribute key instead.

## What?

Introduced backward compatibility to the `input` key if the `pipeline_input` key is missing in the pipeline definition.

Also added a comment that this support for backwards compatibility will be dropped in the next major release of ADF.

---


sbkok:
Pipeline failure issue is resolved with PR #589. So let's merge that one first.

javydekoning (approved):
Thanks for including the tests.

StewartW (approved):

Issue #595, opened 2023-01-27
[Bug]: Management account cannot be managed by ADF's account management


sbkok:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The AccountManagementStateMachine is responsible for creating and updating accounts that are configured in the `adf-accounts` folder of the `aws-deployment-framework-bootstrap` repository.

If that contains a file that instructs ADF to configure the management account, this will fail in the ConfigureAccountAlias step.

### A note about the impact:
The management account cannot be created, obviously, otherwise there wouldn't be an organization where ADF can run from. However, this feature is sometimes used to update the account configuration with ADF.

This issue only impacts ADF installations where the management account is configured in the `adf-accounts` folder.

### Expected Behavior

It should be able to update the account alias of the management account.

### Current Behavior

It fails to determine the account aliases configured for the management account.

The error states that it is missing the `iam:ListAccountAliases` permission for the cross account access role.

### Steps To Reproduce

1. Create an adf-accounts file for the management account.
2. Wait for the AccountManagementStateMachine to process that.

### Possible Solution

Add the missing permission to the `adf-master-account-bootstrap-policy`.

### Additional Information/Context

Please note: I am working on a fix to address this in v3.2.1.

### ADF Version

v3.2.0

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.1](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.1) just now.
I'm hereby closing this issue. Please open a new issue if you are experiencing any issues with the latest release.

PR #596, opened 2023-01-27
Fix management account config alias through ADF account management

sbkok:
Issue: #595

## Why?

The management account alias cannot be configured by the account management state machine, as the cross account access role does not have enough permissions to create, list, or delete the aliases of the management account.

## What?

The cross-account access role that is deployed in the management account gets the required permissions.

These will only be applied if the management account is bootstrapped by ADF.

It assumes into the cross-account access role to list, create, and/or delete the account aliases. Hence, this change removes the permissions from the Lambda function role itself. As those are not used and would only apply to the deployment account.

---


StewartW (approved):
Looks good!

javydekoning (approved):

Issue #597, opened 2023-01-30
[Bug]: AccountManagementStateMachine does not start when creating new account.


PasiPP:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

We updated ADF to 3.2.0 recently and discovered that `AccountManagementStateMachine` and `AccountBootstrappingStateMachine` does not start while creating a new account under `aws-deployment-framework-bootstrap/adf-accounts`. The bootstrap pipeline goes through nicely and flows to deployment account, where other actions are performed accordingly.

State machines (`AccountManagementStateMachine, AccountBootstrappingStateMachine`) in management account are present and active, but they never run. Therefore we can't currently create an account via ADF.

### Expected Behavior

We expect that the bootstrap pipeline triggers the state machines which handles the account creation process on management account.

### Current Behavior

The LogLevel on `serverlessrepo-aws-deployment-framework` was set on DEBUG but it gave no clear reason why the state machine does not start. Everything runs trough successfully pipeline wise and therefore we don't have a clear sight on the possible fix. 

For now, there is nothing that seems to trigger the state machines in the management account.

### ADF Version

3.2.0

StewartW:
Hi, 

Just to confirm, do you have a cloudtrail configured for your management account?

PasiPP:
We encountered the following error during the troubleshoot: 

[...]

PR #598, opened 2023-01-31
Sanitizing account names before using them in SFn Invocation

StewartW:
# Why?

There is a bug, where if an account full name has white space in it, it will not trigger the relevant account management state machine. 

*Issue #, if available:* #597

## What?

Description of changes:
Strip unsupported chars from the account name before starting a state machine execution. 

---

### DIFF
--- a/src/lambda_codebase/account_processing/process_account_files.py
+++ b/src/lambda_codebase/account_processing/process_account_files.py
@@ -42,6 +42,7 @@ class AccountFileData(TypedDict):
     Class used to return YAML account file data and its related
     metadata like the execution_id of the CodePipeline that uploaded it.
     """
+
     content: Any
     execution_id: str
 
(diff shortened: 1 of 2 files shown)


src/lambda_codebase/account_processing/process_account_files.py
```diff
@@ -129,19 +127,52 @@ def process_account(account_lookup, account):
...
+        ",",
+        ";",
+        ":",
+        "/",
+    ]
+    sanitized_name = account_name[:30]
+    for char in unsupported_characters:
+        sanitized_name = sanitized_name.replace(char, "")
+
+    return sanitized_name
```
sbkok, line 173: I think we could change this to a regular expression that will take any non-alphanumeric or underscore char to an underscore. Like so:

```suggestion
def sanitize_account_name_for_snf(account_name):
    return re.sub('[^a-zA-Z0-9_]', '_', account_name[:30])
```

sbkok (changes requested):

src/lambda_codebase/account_processing/process_account_files.py
StewartW: Okay, makes sense. I'll make that change later in the day.

sbkok (approved):
Thank you for fixing this! LGTM.

javydekoning (approved):

PR #606, opened 2023-02-15
Update buildspec.yml

lasv-az:
Avoid build error by upgrading pip

# Why?

Build stage fails due to an old pip version being used. Upgrade it to avoid the stage to fail.

*Issue #, if available:*

## What?

Description of changes:

- List
- What
- You
- Changed

---


javydekoning (approved):

sbkok (changes requested):
Could you please update the specific pip version in the following file instead?

https://github.com/awslabs/aws-deployment-framework/tree/master/src/adf-build/requirements.txt

Otherwise it will downgrade pip right after. Plus the other adf-build/requirements.txt installs will also benefit from this.

Or is there a reason why this wouldn't work?
If we need to update pip before the install requirement, maybe having a two-stage `pip install -r` would be the better approach. What are your thoughts on this?

sbkok (approved):
I made up my mind. Pip will need to be updated before installing the requirements.
So lets keep it as suggested in this PR.

Thanks for contributing this.

PR #623, opened 2023-03-31
Feat/environment agnostic custom images references

abhi1094:
# Why?

As part of our enterprise deployment, we run ADF in three instances across three AWS Organizations (Dev Org, Int Org, Prod Org).
The key goal of this deployment is to have environment agnostic deployment maps. 
A deployment map should be pushed from one org to another with a single merge request. 
This approach only works, when the deployment maps are free of any hardcoded AWS account ids.

To enable this approach, we adjusted the way custom codebuild images are read in from the deployment map. 
With this change, developers can decide to do the following: 
1. Provide the full ARN - as usual. 
2. Provide only the ECS repository name (instead of the ARN). In this case, the account id and region are defaulting to the deployment account and ADF main region.

Furthermore, we fixed a permission issue with custom codebuild images.

## What?

Changed the **determine_build_image** function in adf_codebuild.py to check for image which does not start with arn and then this logic will construct the arn by setting the deployment account id, deployment region and image name. The default behaviour of the code is not changed.

Fixed ecr permissions that helps codebuild to build and push images to the ecr repo.

---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py
```diff
@@ -373,6 +373,20 @@ def determine_build_image(codebuild_id, scope, target, map_params):
...
+                repo_arn = _ecr.Repository.from_repository_arn(
+                    scope,
+                    f'custom_repo_{codebuild_id}',
+                    constructed_repository_arn
+                )
+                return _codebuild.LinuxBuildImage.from_ecr_repository(
+                    repo_arn,
+                    specific_image.get('tag', 'latest'),
+                )
+            # We take the full ECR Arn - Default Behaviour
```
sbkok, line 389: Great addition, I just prefer that we change it slightly.
At the moment, we are using the `repository_arn` as a way to pass the name of the ECR repository.
The name of the property does not make this clear.

I would rather have an additional property, which is `repository_name` or something similar.
If that is specified, you build up the repository arn the way you did.

After constructing the repository arn, could you make sure we reuse the from_repository_arn and from_ecr_repository calls that were there already?

So basically, if the repository_arn property is specified we use that.
Else, if the repository_name is specified we build up the arn that way.
If both are not specified, we raise an error.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py
```diff
@@ -375,15 +375,35 @@ def determine_build_image(codebuild_id, scope, target, map_params):
...
+                    scope,
+                    f'custom_repo_{codebuild_id}',
+                    specific_image.get('repository_arn', ''),
+                )
+                return _codebuild.LinuxBuildImage.from_ecr_repository(
+                    repo_arn,
+                    specific_image.get('tag', 'latest'),
+                )
+            raise Exception(f"Repository arn or Repository name is not specified")
+
```
sbkok, line 406:
```suggestion
            if not repository_arn and not repository_name:
                raise ValueError("The repository arn or name needs to be specified")

            if repository_arn and repository_name:
                raise AssertionError("Specify the arn or the name of the repository, not both.")

            if repository_name:
                repository_arn = (
                    f"arn:aws:ecr:{ADF_DEPLOYMENT_REGION}:"
                    f"{ADF_DEPLOYMENT_ACCOUNT_ID}:{repository_name}"
                )

            ecr_repo = _ecr.Repository.from_repository_arn(
                scope,
                f'custom_repo_{codebuild_id}',
                repository_arn,
            )
            return _codebuild.LinuxBuildImage.from_ecr_repository(
                ecr_repo,
                specific_image.get('tag', 'latest'),
            )

```

sbkok (changes requested):
Could you also update the [CodeBuild properties in the provide guide]](https://github.com/awslabs/aws-deployment-framework/blob/8d0445b1bf5c01f88b920516407a10b0c843c96e/docs/providers-guide.md#properties-4) please?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py
AndyEfaa: Code Looks good to me - @sbkok

sbkok (approved):

AndyEfaa:
Thank you. LGTM.

Issue #627, opened 2023-04-25
There is already a Construct with name 'iam-target-1-wave-0' in Stack [iam]


andrebrc:
We are using deployment maps to create pipelines and when I try to include a new target in some existent pipelines, I've got the error below after the state machine runs the codebuild project:

2023-04-25 12:12:39,443 \| INFO \| __main__ \| ADF Version 3.2.0 \| (generate_pipeline_stacks.py:27)
179 | 2023-04-25 12:12:39,444 \| INFO \| __main__ \| ADF Log Level is INFO \| (generate_pipeline_stacks.py:28)
180 | 2023-04-25 12:12:39,453 \| INFO \| cdk_stacks.main \| Pipeline creation/update of iam commenced \| (main.py:36)
181 | jsii.errors.JavaScriptError:
182 | Error: There is already a Construct with name 'iam-target-1-wave-0' in Stack [iam]
183 | at Node.addChild (/tmp/jsii-kernel-Nkb8Pe/node_modules/constructs/lib/construct.js:458:19)
184 | at new Node (/tmp/jsii-kernel-Nkb8Pe/node_modules/constructs/lib/construct.js:35:28)
185 | at new ConstructNode (/tmp/jsii-kernel-Nkb8Pe/node_modules/@aws-cdk/core/lib/construct-compat.js:174:28)
186 | at Object.createNode (/tmp/jsii-kernel-Nkb8Pe/node_modules/@aws-cdk/core/lib/construct-compat.js:35:42)
187 | at new Construct (/tmp/jsii-kernel-Nkb8Pe/node_modules/constructs/lib/construct.js:503:32)
188 | at new Construct (/tmp/jsii-kernel-Nkb8Pe/node_modules/@aws-cdk/core/lib/construct-compat.js:33:9)
189 | at Kernel._create (/tmp/tmpx_9xgv50/lib/program.js:9964:29)
190 | at Kernel.create (/tmp/tmpx_9xgv50/lib/program.js:9693:29)
191 | at KernelHost.processRequest (/tmp/tmpx_9xgv50/lib/program.js:11544:36)
192 | at KernelHost.run (/tmp/tmpx_9xgv50/lib/program.js:11504:22)

I was thinking of deleting the stacks and re-execute the aws-deployment-framework-pipelines.

Any idea how to solve it?

sbkok:
[...]

Issue #209, opened 2020-01-15
using codebuild as deploy paired with ou path gives unclear error


bundyfx:
```
   default_providers:
      source:
        provider: codecommit
        properties:
          account_id: "123456789012"
      deploy:
        provider: codebuild
        properties:
          image: AMAZON_LINUX_2
          spec_filename: deployspec.yml`
     targets:
         - /ou/test/123  <-- expecting account id
```

Results in an error about construct names for the stage already existing as it attempts to create multiple codebuild stages with the same name. This works as intended when using account ids as opposed to OU paths as it then only targets a single account and thus there is no duplicate stage name attempts.

The question becomes what would we like this to do when using a path? an idea could be that it fetches the account ids from the ou path and inserts them as an environment variable into a single codebuild project (stage).

teemu-laakso:
Hi! Is there any update on this? We ran into the same problem and the proposed solution for adding the target account as an environment variable would work for us. This functionality could be added to any/all pipelines where the deployment is done through codebuild (it would not break the single account use case which already works).

Additionally the region could passed along in a similar manner, and if multiple regions are passed then the result could be creating stages for all combinations:

accounts: [123, 456]
regions: eu-north-1, eu-west-1

combinations: [(123, eu-north-1), (123, eu-west-1), (456, eu-north-1), (456, eu-west-1)]

PR #628, opened 2023-04-28
Fixing codebuild stage naming bug

pozeus:
# Why?

This PR fixes an issue where multiple codebuild stages with the same name were being created when using OU paths. The error occurred because the construct names for the stages already existed, leading to duplicate stage name attempts. This issue did not occur when using account IDs since it targeted only a single account. This PR resolves the issue by ensuring unique construct names are used for each stage, regardless of whether OU paths or account IDs are used.

Resolves:
#627 
#209 

## What?

When using OU paths to create multiple codebuild stages, the current implementation results in an error about duplicate construct names for the stages. This occurs because the construct names for the stages are not unique, leading to multiple attempts to create stages with the same name.

To fix this issue, this PR modifies the code to ensure unique construct names are used for each stage. This is achieved by appending a target Account ID to the construct name based on the stage name and the OU path or account ID. This ensures that each stage has a unique construct name, even if multiple stages have the same name.

The change is backward compatible and does not affect existing codebuild stages. It only ensures that future stages created using OU paths have unique construct names.

---


sbkok (approved):
Thanks for addressing this issue!

StewartW (approved):

PR #633, opened 2023-06-07
Feat/CodeCommit provider - set "default-scm-codecommit-account-id" in adfconfig.yml and set default fallback

AndyEfaa:
# Why?

As part of our enterprise deployment, we run ADF in three instances across three AWS Organizations (Dev Org, Int Org, Prod Org). To allow for easy deployments across those three AWS Organizations, we do need org-agnostic deployment maps. This only works, when the deployment maps are free of any hardcoded AWS account ids.

Therefore, we made the CodeCommit account id in the CodeCommit provider section optional; ensuring deployment maps that are free of hardcoded AWS account ids (at least for this provider).

Our proposed change is already used in an enterprise deployment across three AWS Organizations.

## What?

With this change, the following two features are added: 
1. If not set, the deployment account id is used as the CodeCommit account id in the CodeCommit provider section.
2. Additionally, the CodeCommit account id for the CodeCommit provider section can be set globally, in the adfconfig.yml via the setting: default-scm-codecommit-account-id


src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_or_update_rule.py
```diff
@@ -58,6 +60,37 @@ def lambda_handler(event, _):
...
+        deployment_account_id = DEPLOYMENT_ACCOUNT_ID
+        try:
+            parameter_store = ParameterStore(DEPLOYMENT_ACCOUNT_REGION, boto3)
+            default_scm_codecommit_account_id = parameter_store.fetch_parameter(
+                "/adf/scm/default-scm-codecommit-account-id"
+            )
+        except ParameterNotFoundError:
+            default_scm_codecommit_account_id = deployment_account_id
+        if not source_account_id:
+            print("account_id not found in source_props - recreate it!")
```
StewartW, line 79: Nit: Change this to use a logger.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_repository.py
```diff
@@ -67,6 +68,34 @@ def lambda_handler(event, _):
...
+    #   - set via default_scm_codecommit_account_id (if exists)
+    #   - or set via ADF_DEPLOYMENT_ACCOUNT_ID
+    try:
+        default_scm_codecommit_account_id = parameter_store.fetch_parameter(
+            "/adf/scm/default-scm-codecommit-account-id"
+        )
+    except ParameterNotFoundError:
+        default_scm_codecommit_account_id = DEPLOYMENT_ACCOUNT_ID
+    if not code_account_id:
+        print("account_id not found in source_props - recreate it!")
```
StewartW, line 84: Nit: Don't use print, use a logger

StewartW: Also, no need to yell, can we make this message more.... relevant?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
```diff
@@ -160,6 +160,18 @@ Resources:
...
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Effect: Allow
+                Action:
+                  - "ssm:GetParameter"
+                  - "ssm:GetParameters"
+                  - "ssm:GetParametersByPath"
+                Resource:
+                  - "*"
```
StewartW, line 174: Can this be scoped to the relevant prefix?

StewartW (changes requested):

StewartW:
Hi, thanks for this contribution. 
I really love the idea, think it will be super useful to other customers. 
I've added a few comments to your CR.

Can you add the relevant unit tests please?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_or_update_rule.py
AndyEfaa: Added more verbose comment

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_repository.py
AndyEfaa: Added more verbose comment

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/schema_validation.py
```diff
@@ -333,7 +343,10 @@
 TARGET_SCHEMA = {
     Optional("path"): Or(str, int, TARGET_LIST_SCHEMA),
     Optional("tags"): {
-        And(str, Regex(r"\A.{1,128}\Z")): And(str, Regex(r"\A.{0,256}\Z"))
+        And(str, Regex(r"\A.{1,128}\Z")): Or(
+            And(str, Regex(r"\A.{0,256}\Z")),
+            And(list)
+        )
```
sbkok, line 349: I don't think this change is related to this PR. Can you move that to its own PR if we need to get that changed?

sbkok (changes requested):
First of all, thank you for contributing this!
I would like to request a few changes on the PR.
Please let me know if I need to elaborate or can help with implementing these changes in your PR.

---

At the moment, the same check is performed at multiple places to fallback to the default configured source account id:
1. `src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_or_update_rule.py`
1. `src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_repository.py`
1. `src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codecommit.py`

Ideally, we figure out what value to use once. I would propose moving the check to the logic triggered by this file to be precise:
`src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py`

That builds the input used to build & deploy the pipeline with.
So  instead of storing the variables in the root of the `pipeline_input` object, I would like to propose we check the default source provider here and set the value accordingly.

Next, we can move the `GeneratePipelineInputs` to the start of the pipeline management state machine:
`src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml`

So the `CreateOrUpdateRule` and `CreateRepository` can rely on the correct value from the start.
Instead of having to apply the same logic on both.

Can you look into this?
If you don't have the bandwidth to make this change, I can look into it too if you want.
Please let me know in that case.

---

Separately, I would like to ask if we can add unit tests for the code that is added in this PR.
Can you add those too?

AndyEfaa:
Solution is now simplified based on discussion + tests updated

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_or_update_rule.py
```diff
@@ -58,6 +60,40 @@ def lambda_handler(event, _):
...
+        except ParameterNotFoundError:
+            default_scm_codecommit_account_id = deployment_account_id
+            LOGGER.debug(
+                "default_scm_codecommit_account_id not found in SSM - "
+                "Fall back to deployment_account_id.",
+            )
+        source_account_id = default_scm_codecommit_account_id
+
+        # Create the properties object if it does not exist
+        if not pipeline["default_providers"]["source"].get("properties") is None:
```
sbkok, line 90: I think there is a small merge error here. It states `if not` and `source.properties is None`. In other words, this will evaluate to True if it is set. While the result creates a new properties dictionary.

```suggestion
        if pipeline["default_providers"]["source"].get("properties") is None:
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_repository.py
```diff
@@ -17,7 +17,6 @@
 LOGGER = configure_logger(__name__)
 DEPLOYMENT_ACCOUNT_REGION = os.environ["AWS_REGION"]
 
-
```
sbkok, line 20: According to https://peps.python.org/pep-0008/#blank-lines, there need to be two blank lines before functions start. Can you add this one back?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/schema_validation.py
sbkok: @AndyEfaa Could you move this change to its own PR?
Plus I think it would be good to validate the list of tags too. So instead of the `And(list)` it should validate each item in the list too.

sbkok (changes requested):
Almost there, just a few minor comments

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/create_or_update_rule.py
AndyEfaa: I agree. Changed now

sbkok (approved):
LGTM, thank you for making these changes, this will be very valuable to our end-users!

sbkok (changes requested):
@AndyEfaa A bug slipped into the main branch where failed tests don't stop Tox when it runs into a failure. PR #686 fixes this issue.

If you look at the logs of the ADF CI Tox run, it runs into a failure atm. See `test_prepare_deployment_account_defaults` and `test_prepare_deployment_account_specific_config` in the [logs here](https://github.com/awslabs/aws-deployment-framework/actions/runs/7580710085/job/20646980843?pr=633).

For the moment, if you run tox locally, scroll up to see the red text for now.
This should be resolved in the main branch shortly.

Could you fix those failing tests that occur in this branch?

sbkok (approved):
LGTM, thank you!

StewartW (approved):

PR #634, opened 2023-06-08
feat: allow-for-empty-targets-in-deployment-maps

AndyEfaa:
# Why?

As part of our enterprise deployment, we run ADF in three instances across three AWS Organizations (Dev Org, Int Org, Prod Org). To allow for easy deployments across those three AWS Organizations, we do need org-agnostic deployment maps. This only works, when the deployment maps are identical across all three orgs. 

Even though, the deployment maps are identical across AWS Organizations, the number of created AWS Accounts per AWS org per OU are often not identical. As a result, it happens frequently, that a single OU in a specific AWS Organizations (e.g. Dev Org) has various AWS Accounts mapped, whereas the same OU in the different AWS Organizations (e.g. Int Org) doesn't have any AWS Accounts mapped (yet). Prior to this change, this results in an exception in the AWS Organizations with no AWS Accounts mapped to that OU

To solve for this problem, we introduce the option to allow for deployment maps with "empty target" evaluations. Currently, empty targets result in an exception. With this change, the ADF pipeline still gets created, as defined in the deployment map - just without any targets.

Our proposed change is already used in an enterprise deployment across three AWS Organizations.

## What?

With this change, the following two feature is added:
- In the adfconfig.yml, the optional setting "allow-empty-target" gets introduced. If not set, the default value is "FALSE". That means empty target evaluations still result in an exception, which is the current behavior by ADF prior to this change.

---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -133,8 +143,16 @@ def _create_response_object(self, responses):
...
-            raise NoAccountsFoundError(f"No accounts found in {self.path}")
+
+        if self.allow_empty_deployment_maps == "TRUE":
+            # We now allow empty deployment maps
+            if accounts_found == 0:
+                LOGGER.info(
+                    'Create_response_object: AWS accounts found is 0',
+                )
+        else:
+            if accounts_found == 0:
```
StewartW, line 154: Why not do:
```
if account_found == 0 and self.allow_empty_deployment_maps == "FALSE":
    raise NoAccountsFoundError(....)

```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py
```diff
@@ -140,6 +140,14 @@ def generate_pipeline_inputs(
             list(target_structure.generate_waves()),
         )
 
+    # Print out if final target list is empty...
+    number_targets_total = 0
+    for target in pipeline_object.template_dictionary["targets"]:
+        number_targets_total = number_targets_total + len(target)
```
StewartW, line 146: number_targets_total += len(target)

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py
```diff
@@ -140,6 +140,14 @@ def generate_pipeline_inputs(
...
+    LOGGER.info("We found %s targets total", number_targets_total)
+    if number_targets_total == 0:
+        LOGGER.info("We'll create an empty pipeline with no targets!")
```
StewartW, line 149: Can we change the language here to not log "We"

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -94,6 +99,11 @@ def __init__(
         )
         self.target_structure = target_structure
         self.organizations = organizations
+        parameter_store = ParameterStore(DEPLOYMENT_ACCOUNT_REGION, boto3)
+        self.allow_empty_deployment_maps = parameter_store.fetch_parameter(
```
StewartW, line 103: if you change this to 
```suggestion
        self.allow_empty_deployment_maps = parameter_store.fetch_parameter("/adf/deployment-maps/allow-empty-target") == "TRUE"
```
then future reference to it will be a lot nicer because it's a bool.

StewartW (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py
alexevansigg: number_targets_total = sum([len(target] for target in pipeline_object.template_dictionary["targets"])

src/lambda_codebase/initial_commit/adfconfig.yml.j2
```diff
@@ -27,3 +27,6 @@ config:
   scm:
     auto-create-repositories: enabled
     default-scm-branch: main
+  # deployment-maps:
+    # Optional:
+    # allow-empty-target: "TRUE" # Needs to be "TRUE" or "FALSE". # Defaults to "FALSE", if not set.
```
sbkok, line 32: I suggest we add this with the default value set.

```suggestion
  deployment-maps:
    allow-empty-target: "FALSE"
    # ^ Needs to be "TRUE" or "FALSE". It defaults to "FALSE", if not set. 
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py
```diff
@@ -140,6 +140,14 @@ def generate_pipeline_inputs(
             list(target_structure.generate_waves()),
         )
 
+    # Print out if final target list is empty...
+    number_targets_total = 0
```
sbkok, line 144: I would prefer to rename this variable to: `number_of_targets`.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py
sbkok:
```suggestion
    LOGGER.info("Total number of accounts found: %d", number_targets_total)
    if number_targets_total == 0:
        LOGGER.info("Creating empty pipeline as there were no targets found")
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -133,8 +143,16 @@ def _create_response_object(self, responses):
...
+
+        if self.allow_empty_deployment_maps == "TRUE":
+            # We now allow empty deployment maps
+            if accounts_found == 0:
+                LOGGER.info(
+                    'Create_response_object: AWS accounts found is 0',
+                )
+        else:
+            if accounts_found == 0:
+                raise NoAccountsFoundError(f"No accounts found in {self.path}")
```
sbkok, line 155: To apply what @StewartW commented in a suggestion:
```suggestion
        if accounts_found == 0 and not self.allow_empty_deployment_maps:
            raise NoAccountsFoundError(f"No accounts found in {self.path}")
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -203,9 +221,14 @@ def fetch_accounts_for_target(self):
...
+        if self.allow_empty_deployment_maps == "TRUE":
+            return
+        else: 
+            if self.path is None:
+                # No path/target has been passed, path will default to /deployment
+                return self._target_is_null_path()
+            
+            raise InvalidDeploymentMapError(
+                f"Unknown definition for target: {self.path}"
+            )
```
sbkok, line 234:
```suggestion
        if self.path is None:
            # No path/target has been passed, path will default to /deployment
            return self._target_is_null_path()
            
        raise InvalidDeploymentMapError(
            f"Unknown definition for target: {self.path}"
        )
```

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
AndyEfaa: Changed to bool

sbkok (changes requested):
Thanks for contributing this, I am looking forward to get this merged before the next release if possible.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -94,6 +99,13 @@ def __init__(
         )
         self.target_structure = target_structure
         self.organizations = organizations
+        # Set allow_empty_deployment_maps as bool
+        parameter_store = ParameterStore(DEPLOYMENT_ACCOUNT_REGION, boto3)
+        adf_deployment_maps_allow_empty_bool = parameter_store.fetch_parameter(
+            "/adf/deployment-maps/allow-empty"
```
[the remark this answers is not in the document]
AndyEfaa: `allow-empty-target` is correct. Will change.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -94,6 +99,13 @@ def __init__(
...
+        ).lower().capitalize() == "True"
+        self.allow_empty_deployment_maps = adf_deployment_maps_allow_empty_bool
```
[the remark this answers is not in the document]
AndyEfaa: The goal is to allow for "empty targets for pipelines" - not empty deployment maps.
Changed to `adf_deployment_maps_allow_empty_targets_bool`

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -204,6 +248,10 @@ def fetch_accounts_for_target(self):
             return self._target_is_ou_path(
                 resolve_children=str(self.path).endswith(RECURSIVE_SUFFIX)
             )
+          
+        if self.allow_empty_deployment_maps is True:
```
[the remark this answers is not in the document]
AndyEfaa: Swallowing the error is the correct behavior here in my view.
With this flag set to true, "any" target definition is accepted; Including non-existent targets (in the context of the AWS org in scope).

Example: 
When not including this line, you would get errors as shown below in the case of a non-existent AWS account ids: 
`errors.InvalidDeploymentMapError: Unknown definition for target: 12345678901`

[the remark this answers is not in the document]
sbkok: You are right, this will work correctly.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -251,6 +310,10 @@ def fetch_accounts_for_target(self):
             return self._target_is_ou_path(
                 resolve_children=str(self.path).endswith(RECURSIVE_SUFFIX)
             )
+
+        if self.adf_deployment_maps_allow_empty_target is True:
+            return
+
         if self.path is None:
             # No path/target has been passed, path will default to /deployment
             return self._target_is_null_path()
```
sbkok, line 319:
```suggestion
            self._target_is_null_path()
            return
```

StewartW (approved):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -176,14 +187,32 @@ def _create_response_object(self, responses):
...
+            )
 
     def _target_is_account_id(self):
-        responses = self.organizations.client.describe_account(
-            AccountId=str(self.path)
-        ).get("Account")
-        self._create_response_object([responses])
+        try:
+            responses = self.organizations.client.describe_account(
+                AccountId=str(self.path)
```
sbkok, line 203:
```suggestion
                AccountId=str(self.path),
```

sbkok (approved):
Thanks for contributing this improvement @AndyEfaa!
LGTM.

PR #636, opened 2023-06-08
feat: introduce-variable-org-stage

AndyEfaa:
# Why?

As part of our enterprise deployment, we run ADF in three instances across three AWS Organizations (Dev Org, Int Org, Prod Org). To allow for easy deployments across those three AWS Organizations, we do need org-agnostic deployment maps and buildspec files. 

When doing staged ADF deployments across multiple AWS Organizations, each AWS Organization has certain hard coded values that are specific to that AWS Organization. Furthermore, you might want to run specific tests only in specific AWS Organizations. 

In order to easily establish the AWS Organization stage context in a buildspec, we introduce the new SSM parameter `/adf/org/stage`.

## What?

With this change, the following two features are added: 
1.  in adfconfig.yml, introduce the optional variable `org -> stage`.
2. Create the new SSM parameter `/adf/org/stage` that takes the value from `org -> stage`. Defaults to `none` if `org -> stage` is not set.

---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
StewartW, line 1: Can you update the tests in https://github.com/awslabs/aws-deployment-framework/blob/master/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/tests/test_main.py

sbkok: I've added these

sbkok (approved):
Thank you for contributing this.

sbkok (approved):

StewartW (approved):

Issue #642, opened 2023-07-12
[Bug]: terraform plan doesn't catch errors


igordust:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When using the "plan" stage in terraform pipelines, errors triggered by terraform plan command aren't caught and the step ends wrongfully, stating the success of the step.

### Expected Behavior

Any error happening during terraform plan should interrupt the pipeline execution

### Current Behavior

Adding an error in the terraform code, the terraform plan will succeed no matter what, then it will fail in the terraform apply stage.

### ADF Version

3.2.0

PR #643, opened 2023-07-12
Fix: Added pipefail option so tee is not masking errors anymore

igordust:
# Why?

This fixes issue #642 

## What?

Description of changes:

terraform `plan command` is piped to `tee`, so errors occurring during the plan were masked.
I added pipefail option around terraform plan so that the scripts interrupts the execution and correctly report errors.

---


sbkok (approved):
Thanks for diving into this and fixing the root cause.
Looks good to me!

javydekoning (approved):
Thanks! Good catch, I was looking into this and thought this was already handled by `set -e`, didn't spot the `tee` :)

javydekoning:
Running CI before merge.

javydekoning:
Need to fix #644 , for linter to pass.

PR #645, opened 2023-07-21
Fix GitHub Pipeline secret token usage

sbkok:
**Why?**

No access was provided to pipelines that rely on GitHub repositories as their source, using the GitHub OAuth token stored in the `/adf/github_token` secret in Secrets Manager.

This issue was introduced with the refactoring to the separate pipeline management stack in ADF v3.2.0.

**What?**

Added access rights for the pipeline creation stack to fetch the GitHub OAuth secret.

---


javydekoning (approved):
LGTM!

PR #652, opened 2023-08-08
Fix cleanup of parameters upon pipeline retirement

sbkok:
## Why?

At the moment, the pipeline deletion process attempts to delete the parameters at `/deployment/${pipeline_name}/regions`. However, the path it should use needs to include the source of the pipeline definition. In this case, that is `S3`. So it should have deleted: `/deployment/S3/${pipeline_name}/regions`.

Additionally, if the parameter is not found, it would mark that as a failure. While there is nothing to delete if it isn't found, so it should accept that instead.

## What?

Fixed path and added a catch to make the `ParameterNotFound` exception pass anyway.

---


StewartW (approved):

javydekoning (approved):
TY, looks good to me.

PR #653, opened 2023-08-13
fix: avoid overwritten truncated pipeline definitions

AndyEfaa:
# Why?

The lambda function `DeploymentMapProcessorFunction` is responsible for generating stepfunction executions in `ADFPipelineManagementStateMachine` that create the adf deployment pipelines.

In certain edge-cases, adf deployment pipeline definitions will get skipped / overwritten due to truncated pipeline_names due to the limited space available StepFunction execution Ids. 

This fix solves the problem by generating an new uuid per pipeline_name, if pipeline_names cross a certain length.

More background:
Creation of unique StepFunction execution Ids:
- AWS Step Functions execution id supports max 80 characters.
- However, CodePipeline names can be up to 100 characters.
That means CodePipeline names might need to be truncted. Truncating CodePipeline names
can lead to pipelines having the same name. In that case only the first instance of the 
pipeline with the identical truncated name gets created. The other pipelines doesn't get created
without any error or warning. 
To ensure all defined CodePipelines get created, we check that the truncated pipeline name  
is unique across all deployment maps. If duplicates are detected, we stop using the 
codepipeline_execution_id and generate a fresh one just for this pipeline.
Truncation happens as follows:
- Total max length: stepfunction execution id 80 characters
- full_pipeline_name - Nativly, can be up to 100 characters.
  Gets truncated to 60 chars.
- run_id (uuid) equals 49 characters nativly plus the dash. 
  Gets truncated to 29 chars.

---


AndyEfaa:
@sbkok - all suggested changes are now considered.

sbkok (dismissed):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/process_deployment_map.py
```diff
@@ -173,13 +173,18 @@ def start_executions(
...
+        # hash generated from the pipeline name. If below 80 characters, the
+        # full_pipeline_name + run_id is used.
+        sfn_execution_name_raw = f"{full_pipeline_name}-{run_id}"
+        if len(sfn_execution_name_raw) > 80:
+            truncated_pipeline_name = full_pipeline_name[:60]
+            name_bytes_to_hash = bytes(full_pipeline_name + run_id, 'utf-8')
+            execution_unique_hash = hashlib.md5(name_bytes_to_hash).hexdigest()[:5]
+            sfn_execution_name = f"{truncated_pipeline_name}-{execution_unique_hash}-{run_id}"[:80]
+        else:
+            sfn_execution_name = f"{full_pipeline_name}-{run_id}"
```
StewartW, line 187: nit: 
There's no need for the sfn_execution_name_raw here. 
Why not :
```suggestion
        sfn_execution_name = f"{full_pipeline_name}-{run_id}"
        if len(sfn_execution_name) > 80:
            truncated_pipeline_name = full_pipeline_name[:60]
            name_bytes_to_hash = bytes(full_pipeline_name + run_id, 'utf-8')
            execution_unique_hash = hashlib.md5(name_bytes_to_hash).hexdigest()[:5]
            sfn_execution_name = f"{truncated_pipeline_name}-{execution_unique_hash}-{run_id}"[:80]
```

StewartW (dismissed):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/process_deployment_map.py
```diff
@@ -173,13 +173,16 @@ def start_executions(
...
-        execution_unique_hash = hashlib.md5(name_bytes_to_hash).hexdigest()[:5]
-        sfn_execution_name = f"{truncated_pipeline_name}-{execution_unique_hash}-{run_id}"
+        # characters available. To ensure we don't run over in case of 80+
+        # characters, lets use a truncated version concatenated with an
+        # hash generated from the pipeline name. If below 80 characters, the
+        # full_pipeline_name + run_id is used.
+        sfn_execution_name = f"{full_pipeline_name}-{run_id}"
+        if len(sfn_execution_name) > 80:
+            truncated_pipeline_name = full_pipeline_name[:60]
+            name_bytes_to_hash = bytes(full_pipeline_name + run_id, 'utf-8')
```
sbkok, line 183: Looking at this code again, I think it would be better to only hash the `full_pipeline_name` (without the `run_id`).
As in that case the truncated pipeline name + hash is always consistent. So you know what pipeline to monitor if you are debugging it in multiple runs. As how it is implemented today, the hash would change with each new deployment map change.

I didn't think about this when reviewing the PR that introduced this hashing.
What are your thoughts on this @AndyEfaa?

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/process_deployment_map.py
AndyEfaa: @sbkok - This is fine for me. I just changed it.

sbkok (approved):
LGTM, Thank you!

javydekoning (approved):

PR #657, opened 2023-09-27
Limit adf-state-machine-role to what is needed

alFReD-NSH:
# Why?

Following security best practices and fixing things Aquasec/Cloudsploit is complaining about!

## What?

- Reduced trust policy to only states services as it's only passed to a state machine
- Reduced permissions to only invoke Lambda and publish SNS with only the resources that it uses 
---


StewartW (approved):

sbkok (approved):
Looks great, thank you for improving this!

Issue #664, opened 2023-10-24
[Bug]: Account Based Terraform variables do not work


donnyDonowitz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When deploying ADF with terraform, there are global variables that apply to every account and local variables that apply only to specific accounts. According to the [User Guide](https://github.com/awslabs/aws-deployment-framework/blob/master/docs/user-guide.md#deployment-procedure) this should work with the following folder structure.

```txt
     tfvars <-- This folder contains the structure to define Terraform
     │          variables
     │
     └───global.auto.tfvars <-- this file contains global variables applied to
     │                          all the target accounts
     │
     └───111111111111 <-- this folders contains variable files related to
     │   │                the account
     │   └──────│   local.auto.tfvars <-- this file contains variables related
     │          │                         to the account
```

When this structure is applied, the [adf_terraform.sh](https://github.com/awslabs/aws-deployment-framework/blob/d8087374a066cae0a563e92997b93f417a9a2683/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/adf_terraform.sh) execution fails with:

```sh
Terraform stage: plan
List of target regions: eu-central-1
List of target account: 111111111111
Running terraform plan on account 111111111111 and region eu-central-1
cp: cannot stat '/codebuild/output/src4248460905/src/tfvars/111111111111/*': No such file or directory
Command did not exit successfully bash adf-build/helpers/terraform/adf_terraform.sh exit status 1
Phase complete: BUILD State: FAILED
Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: bash adf-build/helpers/terraform/adf_terraform.sh. Reason: exit status 1
```

### Expected Behavior

According to adf_terraform.sh, the script created a temporary folder in which all the terraform configuration files are copied.
This  should copy the `global.auto.tfvars` and the account based `local.auto.tfvars` .

### Current Behavior

adf_terraform.sh fails in line 16

```sh
cp -R "${CURRENT}/tfvars/${TF_VAR_TARGET_ACCOUNT_ID}"/* "${CURRENT}/tmp/${TF_VAR_TARGET_ACCOUNT_ID}-${AWS_REGION}"
```

The line should copy all contents of the local account directory into a temporary folder but fails because the copy statement does not work with a wildcard.

### Steps To Reproduce

1. Setup ADF Terraform
2. Add a folder for your target AWS Account
[...]

PR #665, opened 2023-10-24
fix: :bug: copy account based variables correctly

donnyDonowitz:
# Why?

Account based variables are not copied to the temporary directory that holds all terraform files. Without this fix, ADF executions fail with 
```sh
cp: cannot stat '/codebuild/output/src4248460905/src/tfvars/111111111111/*': No such file or directory
```

See #664 for a more detailed description.

*Issue #, if available:*
#664 

## What?

Description of changes:

- Use `/.` (works) instead of `/*` (fails) to copy all contents of a directory

---


sbkok (approved):
Thanks for addressing this.
I thought the . after the directory was not adding any value, only referencing to the directory itself, but what do you know.
I tried it on my end, with a simulated environment and your operation works correctly, TIL, thank you!

javydekoning (approved):

PR #667, opened 2023-10-31
Fix: Fix target schema for tags

AndyEfaa:
Fix target schema for tags

# Why?

The schema definition for tags contained a bug that made schema validation fail in certain edge cases - despite providing a valid target.
This PR fixes the issue.

## What?

Description of changes:
- Change made to file `src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/schema_validation.py`
---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/schema_validation.py
```diff
@@ -333,7 +333,10 @@
 TARGET_SCHEMA = {
     Optional("path"): Or(str, int, TARGET_LIST_SCHEMA),
     Optional("tags"): {
-        And(str, Regex(r"\A.{1,128}\Z")): And(str, Regex(r"\A.{0,256}\Z"))
+        And(str, Regex(r"\A.{1,128}\Z")): Or(
+            And(str, Regex(r"\A.{0,256}\Z")),
+            And(list)
```
sbkok, line 338: I think we should also validate the items in this list.
The [Schema Example](https://github.com/keleshev/schema#example) shows this would be possible I think:

```suggestion
            [And(str, Regex(r"\A.{0,256}\Z"))]
```

sbkok (changes requested):

sbkok (approved):
LGTM, thank you!

javydekoning (approved):
Thanks, this will work!

https://docs.aws.amazon.com/resourcegroupstagging/latest/APIReference/API_TagFilter.html#resourcegrouptagging-Type-TagFilter-Values

PR #670, opened 2023-11-24
feat: Adding support for ADF on china partition

Alexpngal:
# Why?

Adapting ADF for china to run on cn-northwest-1
---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/partition.py
danxie1999, line 1: Suggest to add a domain suffix function:

def get_aws_domain(region_name: str) -> str:
    """
    Get AWS domain suffix
    """
    import re
    _partition = get_partition(region_name)
    if re.match("^aws-.*", _partition):
        return "amazonaws.com.{0}".format(_partition.split("-")[-1])
    else:
        return "amazonaws.com"

PR #668, opened 2023-10-31
Feat/multi org support

alexevansigg:
# Why?

Adding support for Multi-Organization 'Enterprise' ADF Installations.

## What?

Description of changes:

- Added New Documentation Section on Multi-Org
- Added adf/org/stage parameter to all managed accounts
- Added method to read organization specific adfconfig.yml files.
-  Updated Sample ADF Baseline Template with Custom Mappings based on organization parameter.

PR #672, opened 2024-01-05
feat: add multi-org generate_params.py

AndyEfaa:
# Why?

When running ADF in a multi-org setup, this feature allows you to set stage-specific files in the repositories' `params` folder. This feature doesn't break existing parameter generation. It just adds a new feature.

Example: 
Let's assume that we have a three AWS Org setup with a dev, int and prod AWS Org. 
That implies that the SSM param "/adf/org/stage" will have one of the following values: [dev, int, prod]. For now, let's assume the value is set to "dev".
Let's further assume that, the `params` folder of our target repository has the following structure: 
- params:
   - global_dev.yml
   - global_int.yml
   - global_prod.yml
   - global.yml

With this feature, the param file global_dev.yml will be selected based on the value "dev" in SSM param "/adf/org/stage".


docs/user-guide.md
```diff
@@ -646,13 +646,16 @@ each account under that OU.
 deployment map as a target. If only the account number is in the deployment map
 the corresponding OU parameter file will not be referenced.
 
+
+
```
sbkok, line 650: Could you remove these empty lines, so there is only one around this code block?

```suggestion
```

docs/user-guide.md
```diff
@@ -712,6 +715,28 @@ the root of the repository.
 *Note:* Currently only Strings type values are supported as parameters to
 CloudFormation templates when deploying via AWS CodePipeline.
 
+**CloudFormation Parameters in a Multi-Organization ADF Setup**
```
sbkok, line 718: Could you change this to a sub section, like so?
That way we can reference it in GitHub issues and other sections later on.

```suggestion
#### CloudFormation Parameters in a Multi-Organization ADF Setup

```

docs/user-guide.md
```diff
@@ -712,6 +715,28 @@ the root of the repository.
...
 
+**CloudFormation Parameters in a Multi-Organization ADF Setup** 
+The CloudFormation Parameter generation feature is fully compatible with 
+the [Multi-Organization ADF Setup](./multi-organization-guide.md) approach. 
+Let's assume that we have a three AWS Org setup with a dev, int and prod 
+AWS Organization. This implies that the SSM param `/adf/org/stage` will have 
+one of the following three values: `[dev, int, prod]`; depending on the AWS 
+Organization you are in. Let's further assume that your application in scope 
+requires AWS Organization specific parameters.In that case, the `params` should 
+have the following content:
```
sbkok, line 726:
```suggestion

For example, in a setup with three AWS Organizations; with a separate
`dev`, an `int`, and a `prod` AWS Organization. This implies that the
SSM parameter `/adf/org/stage` will have one of the following three
values: `dev`, `int`, or `prod`; depending on the AWS organization 
you are in. Let's further assume that your application in scope 
requires AWS Organization specific parameters. In that case,
the `params` folder should  have the following content: 
```

docs/user-guide.md
```diff
@@ -712,6 +715,28 @@ the root of the repository.
...
+params
+└───global_dev.yml
+└───global_int.yml
+└───global_prod.yml
+└───global.yml
+```
+
+When the application gets deployed, ADF will choose the right parameter file 
+based on the value of the SSM prameter "/adf/org/stage".
+
```
sbkok, line 738:
```suggestion
Where it will prefer the AWS Organization specific configuration
`global_${org_stage}` over the `global` parameters in case they both
match the same parameter or tag.
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/generate_params.py
```diff
@@ -299,6 +301,15 @@ def create_parameter_files(self) -> None:
...
                 )
+                # Compare account_region final to global_stage
+                adf_org_stage = ADF_ORG_STAGE # Fetch from Environ for Start
+                current_params = self._merge_params(
+                    Parameters._parse(
+                        params_root_path=self.cwd,
+                        params_filename=f"global_{adf_org_stage}",
+                    ),
+                    current_params
+                )
```
sbkok, line 312: Let's use the global variable directly instead, so we don't need the
local variable and comment here:

```suggestion
                current_params = self._merge_params(
                    Parameters._parse(
                        params_root_path=self.cwd,
                        params_filename=f"global_{ADF_ORG_STAGE}",
                    ),
                    current_params,
                )
```

sbkok (changes requested):
Could you also look into the failing linting issues that MegaLinter highlighted?

docs/user-guide.md
```diff
@@ -712,6 +713,31 @@ the root of the repository.
...
+The CloudFormation Parameter generation feature is fully compatible with 
+the [Multi-Organization ADF Setup](./multi-organization-guide.md) approach. 
+
+For example, in a setup with three AWS Organizations; with a separate
+`dev`, an `int`, and a `prod` AWS Organization. This implies that the
+SSM parameter `/adf/org/stage` will have one of the following three
+values: `dev`, `int`, or `prod`; depending on the AWS organization 
+you are in. Let's further assume that your application in scope 
+requires AWS Organization specific parameters. In that case,
+the `params` folder should  have the following content:
```
sbkok, line 727: Suggestion fixes trailing white spaces linting error:

```suggestion
The CloudFormation Parameter generation feature is fully compatible with
the [Multi-Organization ADF Setup](./multi-organization-guide.md) approach.

For example, in a setup with three AWS Organizations; with a separate
`dev`, an `int`, and a `prod` AWS Organization. This implies that the
SSM parameter `/adf/org/stage` will have one of the following three
values: `dev`, `int`, or `prod`; depending on the AWS organization
you are in. Let's further assume that your application in scope
requires AWS Organization specific parameters. In that case,
the `params` folder should  have the following content:
```

The dead link issue will be resolved if we merge PR #668 first.

sbkok (approved):
LGTM, thank you for addressing the feedback and contributing this feature.

Note for other reviewers, we need to merge #668 before we can merge this one.
That would resolve the dead link issue highlighted by the markdown linter.

Alexpngal (approved):

PR #677, opened 2024-01-12
New install process

sbkok:
## Why?

With the upcoming release of ADF, CDK v2 support is added. The dependencies that go with it, unfortunately increased the deployment size beyond the limit that is supported by the Serverless Application Repository.

Hence a new installation mechanism is required.

## What?

This change proposes to use `make` as the deployment mechanism. I chose `make`, as this eases support queries on deployments of old version and future versions.

For example, by default, the `make` command will install the latest `aws-sam-cli` and `yq` dependencies. However, if these appear to be incompatible, we can instruct to use a specific version of the make install script:

```bash
make update_makefile

make UPDATE_VERSION=make/2.0 update_makefile
```

Similarly, we can instruct using the AWS SAM CLI used at the time the ADF version was developed by running:

```bash
make deps_build
```

In case someone runs into an issue while building or deploying ADF, they can generate a report of their environment with:

```bash
make version_report
```

This generates a report of all the versions of the build environment.

_Long story short:_
The proposed `Makefile` enables various paths to resolve and investigate issues.

## Changes

* Updated the `Makefile` to act as the new deployment mechanism.
* Moved tox related tasks to test and lint the code to `Makefile.tox`. As these need to run in the virtual environment managed by Tox, not the one managed by the new `Makefile`.
* Updated the docs to support ADF administrators installing and updating using the new `Makefile`.
* Tested with v3.2.0 deployments and the upcoming release of ADF.
* Added support to generate version numbers on customized ADF installations.
* Updated the contribution docs to clarify how to deploy and test.
* Added support to update the `Makefile` in the future, such that a deployment issue in a specific ADF version can be fixed by updating the `Makefile`.
* Updated GitHub workflows accordingly, plus added new tests to ensure the dependencies installed by ADF are compatible with each other.
* Added a new step in the GitHub workflow to test a full build of ADF.
* Updated the Bug Issue template to include the version report and to allow the reporter to indicate that they are working on a fix already.

---


Makefile
```diff
@@ -1,28 +1,223 @@
...
+	@echo "$(CLR_GREEN)Update complete$(CLR_END)"
+
+report_makefile_version:
+	@echo "Makefile: v$(MAKEFILE_VERSION) $$(shasum --algorithm 256 Makefile)"
+
+report_versions: report_makefile_version
+	@echo "$(CLR_YELLOW)*** Beginning of ADF Version Report ***$(CLR_END)"
+	@echo "ADF Source version: $(SRC_VERSION)"
+	@echo "ADF $$(cat src/template.yml | grep SemanticVersion | xargs)"
+	@echo ""
```
javydekoning, line 85: Doesn't work on macOS.

Consider:

```sh
	@echo ""
	@echo "Hardware platform: $$(uname --hardware-platform 2>/dev/null || uname -m 2>/dev/null || echo 'n/a')"
	@echo "Kernel name: $$(uname --kernel-name 2>/dev/null || uname -s 2>/dev/null || echo 'n/a')"
	@echo "Kernel release: $$(uname --kernel-release 2>/dev/null || uname -r 2>/dev/null || echo 'n/a')"
	@echo ""
```

Makefile
```diff
@@ -1,28 +1,223 @@
...
+	@echo "Hardware platform: $$(uname --hardware-platform || echo 'n/a')"
+	@echo "Kernel name: $$(uname --kernel-name || echo 'n/a')"
+	@echo "Kernel release: $$(uname --kernel-release || echo 'n/a')"
+	@echo ""
+	@test -e /etc/os-release && echo "OS Release:" && cat /etc/os-release
```
javydekoning, line 90: Errors out for me.

```sh
usage: uname [-amnoprsv]
Kernel name: n/a
uname: illegal option -- -
usage: uname [-amnoprsv]
Kernel release: n/a

make: *** [report_versions] Error 1
```

Try:

```sh
	@echo ""
	@command -v cat > /dev/null 2>&1 && { command -v grep > /dev/null 2>&1 && test -e /etc/os-release && echo "OS Release:" && cat /etc/os-release || echo 'OS Release: n/a'; } || echo 'OS Release: n/a'
	@echo ""
```

docs/installation-guide.md
```diff
@@ -1,91 +1,621 @@
...
+Hence if you upgrade from `2.x` to `3.x`, it is recommended to check the
+[release notes](https://github.com/awslabs/aws-deployment-framework/releases)
+for details and the breaking changes that it introduced.
+
+### 3.7. Deploy ADF
+
+Next, to deploy ADF, run:
+
+```bash
+make deploy
```
javydekoning, line 279: I got stuck at this part because it can't locate the template.yml

```sh
Error: Template file not found at /Users/javy/github/aws-deployment-framework/template.yml
make: *** [sam_deploy] Error 1
``` 

Didn't see any prior errors. Where does it move the `template.yml` from `src`?

javydekoning (changes requested):

Makefile
sbkok: Interesting, so I tried this before by running uname locally, but apparently this used guname.
Which is different from the uname that sh uses. I have a fix ready and will push that shortly.

Makefile
sbkok: Will fix that too, thanks

docs/installation-guide.md
sbkok: Interesting, did you run `make` before? And did that succeed completely?

in the `sam deploy` step, it should automatically use the `.aws-sam/build/template.yaml`.
Can you check if that exists?

Makefile
sbkok: Fixed, can you test this again?

Makefile
sbkok: Fixed, can you test this again?

docs/installation-guide.md
javydekoning: Looks like `make version_number` was failing because I don't have `yq` installed and for some reason it's not installing it. After installing that through brew I get a little further, but still fail on the `make version_number` (not sure why).

```
make: *** [version_number] Error 1
```

I also face intermittent issues with this docker command which is part of the make file:

```sh
docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
```

It randomly hangs, a restart temporarily fixes it. May need some more time with this. Other containers seem to run fine. I do have a modified docker install so it might be a my machine issue.

javydekoning:
Thanks for the support Simon!

I managed to execute from Cloud9. Leaving my notes here:

- Create Cloud9 env.
- Expand EBS volume to 50GB.
- `sudo reboot`
- Add my AWS Credentials `~/.aws/credentials`. Cloud9 will throw a popup asking if it should stop credential management, have it stop managing credentials. 

```sh
git clone https://github.com/sbkok/aws-deployment-framework.git -b "feat/new-install-process"

cd aws-deployment-framework
git remote add upstream https://github.com/awslabs/aws-deployment-framework.git
git fetch upstream --tags
git checkout v3.2.0

RAW_ADF_URL=https://raw.githubusercontent.com/awslabs/aws-deployment-framework
curl -fsSL     "${RAW_ADF_URL}/make/latest/Makefile"     -o ./Makefile.new
mv ./Makefile.new ./Makefile

make clean
make 
make deploy
```

Thanks for adding the GH workflow to test the build. I'll try and add one for macOS at a later stage.

javydekoning (approved):

Makefile
Alexpngal: Tested, It works correctly

Makefile
Alexpngal: Tested, It works correctly for MacOS Monterrey, Intel Chip

CONTRIBUTING.md
```diff
@@ -16,33 +16,50 @@ features.
...
 and [recently closed](https://github.com/awslabs/aws-deployment-framework/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20)
-issues to make sure somebody else hasn't already reported the issue.
+issues to make sure somebody else has not already reported the issue.
 
 Please try to include as much information as you can.
 Details like these are incredibly useful:
 
 - A reproducible test case or series of steps
 - The version of our code being used
+- For deployment related bugs, please include the details returns by the
```
Alexpngal, line 26: For deployment related bugs, please include the details **returned** by the

Alexpngal:
Thanks for the pull request Simon!

I managed to execute the build and deploy successfully from my local computer (MacOS Monterey 12.3.1, Intel processor)

Alexpngal (approved):

PR #678, opened 2024-01-17
Fix initial value type of allow-empty-targets

sbkok:
### Why?

ADF fails to build in the bootstrap pipeline if no `allow-empty-targets` value is configured in the `adfconfig.yml` file.

Root cause, the default value is a boolean False. While the Parameter Store service expects values to be of type string.

### What?

This change set fixes it by converting the default to string when written to Parameter Store.

---


Alexpngal (approved):
LGTM. 
Open question, would it be overkill adding input sanity checks?

javydekoning (approved):

sbkok:
> LGTM. Open question, would it be overkill adding input sanity checks?

Would be good in another PR at some point, I agree.
A similar sanity check to how we verify deployment maps against the schema.

PR #680, opened 2024-01-17
Fix Shared ADF Lambda Layer builds and add move to ARM-64 Lambdas

sbkok:
### Why?

The Lambda Layer that ADF builds included the `bootstrap/adf-build/shared` folder as its source.

Unfortunately, this resulted in 388 MB of unzipped dependencies. While the maximum limit for AWS Lambda Layers is set at 250 MB.

### What?

After going through each of the lambda functions that use this Lambda Layer, I realized that they only require the files located in `bootstrap/adf-build/shared/python`.

Since this implied no longer requiring AWS CDK in the Lambda Layer, I also used this opportunity to move to the ARM-64 Lambdas in ADF.

The new Lambda Layer is 26 MB.

To build ARM-64 on CodeBuild, I've added the Amazon managed container that enables cross-architecture builds on Docker. That will launch and configure Qemu correctly, so ARM-64 Docker images can execute correctly. I've also fixed the build metadata and updated the `sam build` process to use a container instead. As without it, it could not resolve the dependencies required for PyYaml and others specifically for ARM-64.

Lastly, I've updated the shared lambda layer permissions, so we have control in the stack who can access and build on top of it.

---

### DIFF
--- a/linters/custom-adf-dict.txt
+++ b/linters/custom-adf-dict.txt
@@ -4,6 +4,7 @@ adf
 adfconfig
 awscli
 backoff
+binfmt
 bitnami
 boto
 boto3
--- a/src/lambda_codebase/account_bootstrap.py
+++ b/src/lambda_codebase/account_bootstrap.py
@@ -9,20 +9,22 @@ Stack on the target account.
 """
 
 import os
-import boto3
 
+import boto3
 from botocore.exceptions import ClientError
-from logger import configure_logger
+
+# ADF imports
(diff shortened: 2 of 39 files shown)


sbkok:
Rebased + resolved the conflict with recent changes made in `src/template.yml`.

javydekoning (approved):
Thanks for including the ADF prefix!

Alexpngal (approved):

PR #681, opened 2024-01-17
Add AccessControl on S3 Buckets

sbkok:
### Why?

At the moment, if there is an `AccessControl` set on the S3 Bucket (in v3.2.0 ADF had this set), you cannot remove it and move to the new Bucket Ownership Rules.

The only option to move to the new Bucket Ownership Rules would be to add those and remove the `AccessControl` property after. However, since we need to allow customers to upgrade, we cannot do this in a single release.

### What?

Added the Bucket Ownership Rules today, and keep the `AccessControl` for now.

We can remove the `AccessControl` property in a future version. While we instruct customers to first upgrade to v4.0.0 if they need to move to the version where this property is removed.

---

### DIFF
--- a/.cfnlintrc
+++ b/.cfnlintrc
@@ -15,3 +15,4 @@ include_checks:
   - I
 ignore_checks:
   - W3002
+  - W3045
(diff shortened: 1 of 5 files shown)


.cfnlintrc
javydekoning, line 1: Unfortunately there is no better way right now! Thanks Simon!

javydekoning (approved):

.cfnlintrc
sbkok: I agree, I wish there was an alternative to this issue.

PR #682, opened 2024-01-17
Auto generate unique branch names on new version deployments

sbkok:
### Why?

In case a previous deployment of ADF failed for whatever reason, running it again would require you to delete the branch it created the first time.

### What?

Since we don't want to end up deleting a branch that we did not create, the proposed solution is to generate a unique name for the branch and use that.

1. It will start with the version number as the branch name, like `v3.2.0`.
2. If that exists, it will iterate from `v3.2.0-no-1` to `v3.2.0-no-9`.
3. Highly unlikely, but if all of these exist, it will fallback to `v3.2.0-at-{unix_timestamp}`.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/initial_commit/initial_commit.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/initial_commit/initial_commit.py
@@ -9,6 +9,7 @@ from typing import Mapping, Optional, Union, List, Dict, Any, Tuple
 from dataclasses import dataclass, fields
 from enum import Enum
 from pathlib import Path
+from datetime import datetime, timezone
 import re
 import boto3
 import jinja2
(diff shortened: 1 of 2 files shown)


Alexpngal:
Thank you Simon for this contribution!! I agree with Javy on having the commit hash as main identifier alongside the adf version.

sbkok:
That is generated by the new installation process already.
If you deploy a stock version, like `v3.2.0`, it will; generate the ADF version number `v3.2.0`.
The branch name would be: `v3.2.0`, or if needed `v3.2.0-no-1` to  `v3.2.0-no-9`, etc.

If you made changes, and created a commit, it will generate the ADF version number as: `v3.2.0-1-abcdef01`.
The branch name would be: `v3.2.0-1-abcdef01`, or if needed `v3.2.0-1-abcdef01-no-1` to `v3.2.0-1-abcdef01-no-9`, etc.

The generated version number is:
```typescript
`${latest_tag_version_that_is_in_the_commit_chain}-${number_of_commits_since_tag}-${short_commit_id}`
```

Alexpngal (approved):

PR #685, opened 2024-01-18
New feature: terraform target by tags

igordust:
# Why?

With this new feature, terraform helper is able to target accounts matching a tag expressed as key=value. You can define multiple tags and all accounts matching at least one of the tags will become target of the pipeline.

## What?

Description of changes:

- Added proper documentation for the new parameter TARGET_TAGS
- Added the functions needed to fetch tags from accounts in AWS Organization and produce the output file in `get_accounts.py`
- Added `organizations:ListTagsForResource`permission to `adf-organizations-readonly-policy` so that the pipeline role is able to fetch tag information
- Added the logic in adf_terraform.sh to read the list of accounts matched by tags.

---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -38,6 +39,12 @@ def main():
         with open('accounts_from_ous.json', 'w', encoding='utf-8') as outfile:
             json.dump(accounts_from_ous, outfile)
 
+    if TARGET_TAGS:
+        print("filtering by tags")
```
sbkok, line 43: Could you remove this or change it to a `LOGGER` statement?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -90,6 +97,45 @@ def get_accounts():
...
+        'organizations',
+        (
+            f'arn:{PARTITION}:sts::{MANAGEMENT_ACCOUNT_ID}:role/'
+            f'{CROSS_ACCOUNT_ACCESS_ROLE}-readonly'
+        ),
+        'getaccountIDs',
+    )
+    filtered_accounts = []
+    for account in accounts:
+        tags = account['Tags'] = organizations.list_tags_for_resource(
```
sbkok, line 125: Instead of `list_tags_for_resource` and filtering here, you could also rely on the implementation that ADF uses for CodePipeline targets:

* https://github.com/awslabs/aws-deployment-framework/blob/f2bb3471c1c56a8fd42a33a284f50e98ba47a12e/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py#L423-L429
* https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/resourcegroupstaggingapi/client/get_resources.html

Ideally, the way the targets are defined would match the same style as defined in the deployment map schema.
Which also allows a single tag with multiple values. Where the tag key should match any of that list.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -90,6 +97,45 @@ def get_accounts():
...
+            ResourceId=account['AccountId']
+        )['Tags']
+        for tag_filter in tag_filters:
+            found = list(filter(lambda item: (
+                item["Key"] == tag_filter["Key"] and item["Value"] == tag_filter["Value"]), tags))
+            if len(found) > 0:
+                print(
```
sbkok, line 132: Please remove this `print` or replace by a `LOGGER` statement.

docs/user-guide.md
```diff
@@ -1109,6 +1109,11 @@ pipeline for more details in the setup and integration.
 - `TARGET_ACCOUNTS`: comma separated list of target accounts.
 - `TARGET_OUS`: comma separated list of target leaf OUs (parent
   OUs are supported).
+- `TARGET_TAGS`: comma separated list of tag=value couples. All accounts
+  matching at least one of the tags will be targeted by the pipeline.
+  
+  **Eg:** `TARGET_TAGS=environment=prd,cost-center=ccoe` will match all
+  accounts tagged with `environment = prd` OR `cost-center = ccoe`
```
sbkok, line 1116: Tag keys and values are also allowed to have a `=` key. See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

It looks like a comma is not allowed, so that should be good.
Is there another way to pass the target_tags?

sbkok (changes requested):
Thanks for contributing this feature, would be very useful to have!

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
igordust: Resolved with commit 646a3b485afe3fb56af23d20a8ccecca259b846b

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
igordust: Resolved with commit 646a3b485afe3fb56af23d20a8ccecca259b846b

docs/user-guide.md
igordust: Updated the documentation with the new filter format, see 217e71736fd930b56957024face541f78227cc67

docs/user-guide.md
```diff
@@ -1109,6 +1109,16 @@ pipeline for more details in the setup and integration.
...
+- `TARGET_TAGS`: specify filter tags with the following format:
+  ```Name=tagName1,Values=value1,value2,val*;Name=tagName2,Values=value3,value4```
+
+  Each clause is composed by `Name` and `Values` fields. You can input multiple values in a clause, separated
+  by comma, use `*` wildcard to match zero or more characters, use `?` wildcard to match exactly a character
+
+  All filter clauses are applied with logical AND, all values in a single clause are applied with logical OR
+
+  **Eg:** `TARGET_TAGS=Name=environment,Values=prd,dev;Name=cost-center,Values=ccoe` will match all
+  accounts tagged with `environment=prd` OR `environment=dev` AND `cost-center = ccoe`
```
sbkok, line 1121:
```suggestion
  accounts tagged with (`environment=prd` OR `environment=dev`) AND `cost-center=ccoe`
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -90,6 +98,41 @@ def get_accounts():
...
+        tag_values = tags.split(",", 1)[1].split("=")[1].split(",")
+        tag_filters.append({
+            "Key": tag_name,
+            "Values": tag_values})
+    LOGGER.info(
+        "Tag filters %s",
+        tag_filters
+    )
+    organization_api_region = get_organization_api_region(REGION_DEFAULT)
+    print(organization_api_region)
```
sbkok, line 114:
```suggestion
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -90,6 +98,41 @@ def get_accounts():
...
 
 
+def get_accounts_from_tags():
+    tag_filters = []
+    for tags in TARGET_TAGS.split(";"):
+        tag_name = tags.split(",", 1)[0].split("=")[1]
+        tag_values = tags.split(",", 1)[1].split("=")[1].split(",")
+        tag_filters.append({
+            "Key": tag_name,
+            "Values": tag_values})
```
sbkok, line 108:
```suggestion
            "Values": tag_values,
        })
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -90,6 +98,41 @@ def get_accounts():
...
+        "Tag filters %s",
+        tag_filters
+    )
+    organization_api_region = get_organization_api_region(REGION_DEFAULT)
+    print(organization_api_region)
+    tags_client = get_boto3_client(
+        'resourcegroupstaggingapi',
+        (
+            f'arn:{PARTITION}:sts::{MANAGEMENT_ACCOUNT_ID}:role/'
+            f'{CROSS_ACCOUNT_ACCESS_ROLE}-readonly'
```
sbkok, line 119: This will need to be updated to the new role added in v4.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/terraform/get_accounts.py
```diff
@@ -20,10 +21,12 @@
 
 MANAGEMENT_ACCOUNT_ID = os.environ["MANAGEMENT_ACCOUNT_ID"]
 TARGET_OUS = os.environ.get("TARGET_OUS")
+TARGET_TAGS = os.environ.get("TARGET_TAGS")
 REGION_DEFAULT = os.environ["AWS_REGION"]
 PARTITION = get_partition(REGION_DEFAULT)
 sts = boto3.client('sts')
 ssm = boto3.client('ssm')
+organizations = boto3.client('organizations')
```
sbkok, line 29: Could you change this to include a higher retry count?
Otherwise it might run into the rate limitation of the AWS Organizations API and fail.
This would be likely if multiple Terraform pipelines would run concurrently.

An example how to change it: https://github.com/awslabs/aws-deployment-framework/pull/774/files#diff-05a19c0ecc528b81f98611295fccbb53ec8ce9f20937f67308f3bc886e4a5018R83-R89

sbkok (changes requested):

Issue #659, opened 2023-09-28
[Bug]: Parameter "deployment_account_id" into deployment account's parameter store 


nicobarto95:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The second step function, during ADF bootstrap processes, returns an error because it don't create `deployment_account_id` parameter into parameter store on the deployment account regions target and account region.

### Expected Behavior

Automatic creation of the parameter `deployment_account_id` into parameter store on the deployment account region and the target ones.

### Current Behavior

`{
  "error": "GenericAccountConfigureError",
  "cause": {
    "errorMessage": "An error occurred (ValidationError) when calling the CreateChangeSet operation: Unable to fetch parameters [deployment_account_id] from parameter store for this account.",
    "errorType": "GenericAccountConfigureError",
    "requestId": "e621c29b-9109-4e28-b1fb-4885269c9eaf",
    "stackTrace": [
      "  File \"/var/task/account_bootstrap.py\", line 117, in lambda_handler\n    return _lambda_handler(event, context)\n",
      "  File \"/var/task/account_bootstrap.py\", line 192, in _lambda_handler\n    cloudformation.create_iam_stack()\n",
      "  File \"/opt/python/cloudformation.py\", line 453, in create_iam_stack\n    create_change_set = self._create_change_set()\n",
      "  File \"/opt/python/cloudformation.py\", line 358, in _create_change_set\n    raise GenericAccountConfigureError(error) from error\n"
    ]
  }
}`

### Possible Solution

The solution is manually creation of the `deployment_account_id` into parameter store on the deployment account region and the target ones.

### ADF Version

3.2.0

javydekoning:
Fixed in current master branch. Ref: #695 will be part of 4.0.0.

sbkok:
Thank you for your patience. I am happy to inform you that this issue has been [resolved in our latest release v3.2.1](https://github.com/awslabs/aws-deployment-framework/releases/tag/v3.2.1) just now.
Please open a new issue if you are experiencing any issues with the latest release.

Issue #594, opened 2023-01-27
[Bug]: Fresh installation of v3.2.0 fails to bootstrap the deployment account (upgrades work)


sbkok:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When ADF v3.2.0 is installed fresh, in a new account that did not have ADF installed, it will create a new `global-iam.yml` file in the `adf-bootstrap` folder of the `aws-deployment-framework-bootstrap` repository.

This template, however, requires the `deployment_account_id` parameter to be present in the account. For the deployment account itself, it is not present.

### Expected Behavior

The deployment account should be bootstrapped correctly too.

### Current Behavior

In the management account, the AccountBootstrappingStateMachine in us-east-1 fails to bootstrap the deployment account.

The step that fails is: `CreateOrUpdateBaseStack`.

The error it throws:
```json
{
  "error": "GenericAccountConfigureError",
  "cause": {
    "errorMessage": "An error occurred (ValidationError) when calling the CreateChangeSet operation: Unable to fetch parameters [deployment_account_id] from parameter store for this account.",
    "errorType": "GenericAccountConfigureError",
    "requestId": "12345678-9abc-def0-1234-56789abcdef0",
    "stackTrace": [
      "  File \"/var/task/account_bootstrap.py\", line 117, in lambda_handler\n    return _lambda_handler(event, context)\n",
      "  File \"/var/task/account_bootstrap.py\", line 192, in _lambda_handler\n    cloudformation.create_iam_stack()\n",
      "  File \"/opt/python/cloudformation.py\", line 453, in create_iam_stack\n    create_change_set = self._create_change_set()\n",
      "  File \"/opt/python/cloudformation.py\", line 358, in _create_change_set\n    raise GenericAccountConfigureError(error) from error\n"
    ]
  }
}
```

### Steps To Reproduce

1. Install ADF in a new AWS organization where ADF was not installed before.
2. Wait until it kicks-off the deployment account bootstrapping process, see the AccountBootstrappingStateMachine.

### Possible Solution

The `deployment_account_id` should be created in the deployment account too.
Such that the same template can be applied to all accounts.

Additionally, we need to create a specific `global-iam.yml` file for the deployment ou as well.
In the same way that happens for the root one. Copying the `example-global-iam.yml` to the `global-iam.yml`.
At the moment this is not configured, however, that implies that the global one gets applied in the deployment account too.
[...]

PR #686, opened 2024-01-18
Fix missing deployment_account_id and initial deployment global IAM bootstrap

sbkok:
### Why?

Issues: #659 and #594.

When installing ADF the first time, the global IAM bootstrap stack that gets deployed is sourced from the `adf-bootstrap/global-iam.yml`.

The reason for this behaviour is the absence of the `global-iam.yml` file in the deployment OU bootstrap folder
(`adf-bootstrap/deployment/global-iam.yml`).

It iterates to the parent directory until it finds a `global-iam.yml` to deploy. Hence, when the `adf-bootstrap/global-iam.yml` gets deployed in the deployment account, it was looking for the `deployment_account_id` SSM parameter. That did not get deployed in the deployment account.

### What?

* Add the creation of the `deployment_account_id` in the deployment account, so if the global IAM bootstrap stack failed to deploy before, it will work in the next release. This would be the case if the previous deployment failed but the same `aws-deployment-framework-bootstrap` repository is used in the upgrade.
* When installing the first time, it creates the bootstrap repository. At the time of creation, it will copy the `adf-bootstrap/deployment/example-global-iam.yml` to `adf-bootstrap/deployment/global-iam.yml`. The same logic as how ADF creates the initial `adf-bootstrap/global-iam.yml`.

---

### DIFF
--- a/Makefile.tox
+++ b/Makefile.tox
@@ -11,10 +11,8 @@ all: test lint
 
 test:
 	# Run unit tests
-	( \
-		for config in $(TEST_CONFIGS); do \
-			pytest $$(dirname $$config) -vvv -s -c $$config; \
-		done \
+	@ $(foreach config,$(TEST_CONFIGS), \
+		pytest $$(dirname $(config)) -vvv -s -c $(config) || exit 1; \
 	)
 
 lint:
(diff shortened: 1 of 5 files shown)


javydekoning (approved):

javydekoning (approved):

sbkok:
I moved this change to a larger PR with related changes to ADF's use of SSM Parameter Store.
Closing this one, as #695 supersedes it.

PR #691, opened 2024-02-22
Fix PipelineDeletionStateMachine failing on deleted stacks

alFReD-NSH:
PipelineDeletionStateMachine's IAM role only allowed deleting stacks with a createdBy tag, however non-existing/deleted stacks don't have tags from IAM point of view.

When PipelineDeletionStateMachine is run for a stack that is already deleted but parameter is not deleted, it will try to delete the stack but fails with AccessDenied.

The stack can be deleted for example due to manual deletion or something going wrong when deleting the parameter.

Removing the condition fixes this and DeleteStack will be a noop when stack doesn't exist.

Something to keep in mind is that the protection for this to fail on stacks that are not created by ADF goes away and to keep we will have to complicated the logic to check for stack existance another way.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml
@@ -373,9 +373,6 @@ Resources:
                   - cloudformation:DeleteStack
                 Resource:
                   - "*"
-                Condition:
-                  StringEquals:
-                    'aws:ResourceTag/createdBy': "ADF"
               - Effect: Allow
                 Action:
                   - "ssm:DeleteParameter"


sbkok:
Hi @alFReD-NSH, thanks for highlighting this issue.
Instead of removing the protection, I would propose we check if the stack exists and only try to delete it if it does.
This would also resolve the root cause of this issue I think.

Would that be something you can contribute in this PR instead?

javydekoning:
+1 to what Simon mentioned. There already is a fallback present for the SSM Parameter not existing, but it doesn't exist for Cfn: https://github.com/awslabs/aws-deployment-framework/blob/4e33aba355a30db34000d53b02b043312fb0c68a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/pipeline_management.yml#L617-L631

sbkok:
I'm closing this pull request for now, because it has been inactive for too long.

Feel free to reopen it though :) In case you want to reopen, could you update the pull request as suggested above, rebase your branch, and/or clarify the changes made? Please let us know if you need help with this.

Issue #263, opened 2020-05-20
adf-accounts: Should it create new OUs if they don't exists?


StewartW:
ADF Version: 3.0.1
`
- account_full_name: "some-random-account"
--
  | organizational_unit_path:  /existing/new-ou
  | email:  some-random-account@accounts.skyscnr.com
  | allow_billing: True
  | delete_default_vpc: False
  | alias: some-random-account
  | tags:
  | - managed_by_adf: True

`

In this kind of example, when adding a new account via adf-accounts/<filename>.yml 

It will fail with a ValueError because it can't find the OU. 
[File Here](https://github.com/awslabs/aws-deployment-framework/blob/master/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py#L291)

Do we think this should create the relevant OU if it doesn't exist? 
What's the thoughts on that? 
There's pros: 
- allows for smoother creation of OUs and moving accounts "as code" 
- keeps everything in the same process 

Cons: 
- Could hide errors if there's a typo in an OU name

PR #693, opened 2024-03-14
Feature createorganizationalunit

ethanBaird:
# Why?

In our existing Account Creation workflow, we are required to create new organizational units via the Organizations console in our master account. This requires us to sign in to our master account with heightened priveleges for every new account (that has a new path)

We run a multi account strategy so this happens reasonably often.

We'd prefer for this to be automated. This change will programmatically create any new Organizational Units if a new path is defined in the `adf-accounts` configuration.

Response to Issue: (https://github.com/awslabs/aws-deployment-framework/issues/263)

## What?

Description of changes:

- Updated permissions on `adf-codebuild-role` in master account
- Added a new method `Organizations.create_ou()`
- Refactored existing method `Organizations.get_ou_id()`
- - refactor `while` loop to `for` loop
- - refactor logic in for/else to create OU if not found as defined by the path
- stubbed tests for both methods

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/test_organizations.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/test_organizations.py
@@ -7,12 +7,13 @@ from datetime import datetime, timezone
 import os
 import boto3
 
-from pytest import fixture
+from pytest import fixture, raises
 from stubs import stub_organizations
 from mock import Mock, patch
 from cache import Cache
 from organizations import Organizations, OrganizationsException
 from botocore.stub import Stubber
+from botocore.exceptions import ClientError
 import unittest
 
 
@@ -29,6 +30,12 @@ def test_create_ou(cls):
 
     assert ou['OrganizationalUnit']["Id"] == "new_ou_id"
     assert ou['OrganizationalUnit']["Name"] == "new_ou_name"
+
+def test_create_ou_throws_client_error(cls):
+    cls.client = Mock()
+    cls.client.create_organizational_unit.side_effect = ClientError(operation_name='test', error_response={'Error': {'Code': 'Test', 'Message': 'Test Message'}})
+    with raises(OrganizationsException): 
+        cls.create_ou("some_parent_id", "some_ou_name")
 
 
 def test_get_ou_id_can_create_ou_one_layer(cls):


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,22 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
+            for org_unit in org_units:
+                if org_unit["Name"] == ou:
+                    parent_ou_id = org_unit["Id"]
                     break
             else:
-                raise ValueError(
-                    f"Could not find ou with name {ou_hierarchy} in OU list {org_units}.",
-                )
+                LOGGER.info(f'No OU found with name {ou} and parent {parent_ou_id}, will create.')
+                new_ou = self.create_ou(parent_ou_id, ou)
```
StewartW, line 493: nit: trailing whitespace here

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,22 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
-                )
+                LOGGER.info(f'No OU found with name {ou} and parent {parent_ou_id}, will create.')
+                new_ou = self.create_ou(parent_ou_id, ou) 
+                parent_ou_id = new_ou['OrganizationalUnit']['Id']
 
         return parent_ou_id
+    
+
+    def create_ou(self, parent_ou_id, name):
+        try:
```
StewartW, line 500: nit: whitespace after the `:`

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,22 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
             else:
-                raise ValueError(
-                    f"Could not find ou with name {ou_hierarchy} in OU list {org_units}.",
-                )
+                LOGGER.info(f'No OU found with name {ou} and parent {parent_ou_id}, will create.')
+                new_ou = self.create_ou(parent_ou_id, ou) 
+                parent_ou_id = new_ou['OrganizationalUnit']['Id']
 
         return parent_ou_id
+
```
StewartW, line 497: nit: whitespace and an extra line here.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,22 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
+    def create_ou(self, parent_ou_id, name):
+        try: 
+            ou = self.client.create_organizational_unit(
+                ParentId=parent_ou_id,
+                Name=name
+                )
+        except ClientError as error:
+            LOGGER.error(f'Failed to create OU called {name}, with parent {parent_ou_id}')
+            raise error
```
StewartW, line 507: might be better to raise an OrganizationsException from error here
maybe use logger.exception(error) as well.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,22 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
         return parent_ou_id
+    
+
+    def create_ou(self, parent_ou_id, name):
+        try: 
+            ou = self.client.create_organizational_unit(
+                ParentId=parent_ou_id,
+                Name=name
+                )
+        except ClientError as error:
```
StewartW, line 505: This branch doesn't have a test, could you add on please

StewartW (changes requested):
Thanks for contributing. A few nits and comments but otherwise a great addition. 

The only question I have is if there should be a flag around this functionality to let customers keep the existing behavior.

javydekoning:
@ethanBaird , can you please fix these linter findings?

```sh
  src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py:
  	493: Trailing whitespace
  	497: Trailing whitespace
  	500: Trailing whitespace
  src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/tests/stubs/stub_organizations.py:
  	Wrong line endings or no final newline
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
ethanBaird: Is there something specific missing? I added tests to `tests/test_organizations.py` for this new method.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,23 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
 
+    def create_ou(self, parent_ou_id, name):
+        try:
+            ou = self.client.create_organizational_unit(
+                ParentId=parent_ou_id,
+                Name=name
+                )
+        except:
+            LOGGER.excpetion(f'Failed to create OU called {name}, with parent {parent_ou_id}')
+            raise OrganizationsException()
```
ethanBaird, line 506: @StewartW not sure I've implemented the OrganizationsException properly here. Should I pass the error message to the Exception class rather than the logger?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
ethanBaird: Pushed a change for this below

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -479,23 +479,34 @@ def get_ou_id(self, ou_path, parent_ou_id=None):
...
+
+            for org_unit in org_units:
+                if org_unit["Name"] == ou:
+                    parent_ou_id = org_unit["Id"]
                     break
             else:
-                raise ValueError(
-                    f"Could not find ou with name {ou_hierarchy} in OU list {org_units}.",
-                )
+                LOGGER.info(f'No OU found with name {ou} and parent {parent_ou_id}, will create.')
```
javydekoning, line 492: Use lazy % formatting in logging functions. See details here: https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles

And in the linter:

```sh
============================== 17 passed in 0.30s ==============================
# Linter performs static analysis to catch latent bugs
find ./src -iname "*.py" | xargs pylint --verbose --rcfile .pylintrc
Using config file .pylintrc
************* Module bootstrap_repository.adf-build.shared.python.organizations
src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py:492:16: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)
src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py:506:12: W0707: Consider explicitly re-raising using 'except Exception as exc' and 'raise OrganizationsException() from exc' (raise-missing-from)
```

The same file has other `LOGGER` examples that use lazy % formatting.

You can also run the linter locally to catch these errors. Review: ============================== 17 passed in 0.30s ==============================
# Linter performs static analysis to catch latent bugs
find ./src -iname "*.py" | xargs pylint --verbose --rcfile .pylintrc
Using config file .pylintrc
************* Module bootstrap_repository.adf-build.shared.python.organizations
src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py:492:16: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)
src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py:506:12: W0707: Consider explicitly re-raising using 'except Exception as exc' and 'raise OrganizationsException() from exc' (raise-missing-from)

javydekoning (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
ethanBaird: Thanks for this. Have been able to run this locally now and seems to be passing.

Unsure about the implementation about the exception which I've asked @StewartW for input on.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
StewartW: Nit: typo on the logger.exception call. 
This should look like: 

```
except ClientError as client_err:
     LOGGER.exception("Failed.....", client_err)
     raise OrganizationsException("Failed......") from client_err
```

ethanBaird:
Hey @StewartW  👋 

Sorry, been a while since I've looked at this, but I have updated the `create_ou` method to throw an Organizations exception when encountering a boto `ClientError` and tested the method for this. 🚀 

Regarding making this an optional flag in the `adf_config.yml`. I'm not sure what the best way to do this is.

I understand that we have the SSM Parameter `config` which we can read in. But should we be passing an `ssm_client` to every `Organizations()` object in order to read this?

Something like... in `organizations.py`

```python
self.config = ssm_client.get_parameter('config')
```

Or I see we use the `ParameterStore` class in a few instances. Is it better to pass this to the Org Class and so something more like

```python
self.config = parameter_store.fetch_parameter('config')
```

Let me know what you think.

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
ethanBaird: Fixed 🚀

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
ethanBaird: Added test.

StewartW:
> self.config = parameter_store.fetch_parameter('config')

This is the better approach yeah.

PR #695, opened 2024-03-26
Add /adf params prefix and other SSM Parameter improvements

sbkok:
## Why?

Issues: #659 and #594.

When installing ADF the first time, the global IAM bootstrap stack that gets
deployed is sourced from the `adf-bootstrap/global-iam.yml`.

The reason for this behaviour is the absence of the `global-iam.yml` file
in the deployment OU bootstrap folder
(`adf-bootstrap/deployment/global-iam.yml`).

It iterates to the parent directory until it finds a `global-iam.yml` to
deploy. Hence, when the `adf-bootstrap/global-iam.yml` gets deployed in the
deployment account, it was looking for the `deployment_account_id` SSM
parameter. That did not get deployed in the deployment account.

---

Pytest failures were ignored due to a change in the Makefile
used to execute tests. The ADF CI GitHub Workflow would result in a
success, even when a test case failed.

---

Some of the parameters ADF created would be placed in the
root of the SSM Parameter Store while others were neatly located in the `/adf` path.

## What?

* Add the creation of the `deployment_account_id` in the deployment account,
  so if the global IAM bootstrap stack failed to deploy before, it will work
  in the next release. This would be the case if the previous deployment failed
  but the same `aws-deployment-framework-bootstrap` repository is used in the
  upgrade.
* When installing the first time, it creates the bootstrap repository.
  At the time of creation, it will copy the
  `adf-bootstrap/deployment/example-global-iam.yml` to
  `adf-bootstrap/deployment/global-iam.yml`.
  The same logic as how ADF creates the initial `adf-bootstrap/global-iam.yml`.
  
---

Fixed by exiting on the first failure using Makefile foreach instead.

---

Add a `/adf/` prefix to parameter names to ease access management and making
it easier to distinguish ADF parameters from other solutions.

To enable upgrades, the account handler function that performs the lookup
or creation of the deployment account is updated to rely on the
AWS Organizations API to check if there are any deployment accounts in the
`/deployment` organization unit path.

Upon an update, it will use the AWS account if only one is in that specific OU.
If there are more, it will error and instruct the user to move unnecessary
accounts out of the `/deployment` organization unit first and try again.

---

Refactor master references to management or main.

---

Fix default_scm_codecommit_account_id and other scm related params under the /adf/scm path. Changed param name to underscores for consistency.

---


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_github.py
```diff
@@ -53,7 +53,7 @@ def create_webhook_when_required(scope, pipeline, map_params):
                 .get('properties', {})
                 .get('branch')
             )
-            or 'master'
```
StewartW, line 56: Does this need to be backwards compatible ```... or 'main' or 'master'```

StewartW (approved):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_github.py
sbkok: With v4.0.0 the default branch will change to 'main'.
I will add a CHANGELOG.md file describing these breaking changes shortly. So we can write-up the breaking changes ahead of the release.

javydekoning (approved):

PR #700, opened 2024-03-28
Fix tagging of Cross Account Access role in the management account

sbkok:
## Why?

When ADF needs to tag or untag the cross-account access role, it fails to do so.

## What?

* Granting the required permissions to tag and untag the roles.
* Removal of the invalid `s3:ListObjects` permission, the permissions required to list objects in a bucket are managed via the `s3:ListBucket` permission. Which is already present.

---

### DIFF
--- a/src/template.yml
+++ b/src/template.yml
@@ -1307,7 +1307,6 @@ Resources:
               - "s3:GetObject"
               - "s3:ListBucket"
               - "s3:PutObject"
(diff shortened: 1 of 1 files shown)


javydekoning (approved):
Oops, I was under the impression we validate IAM policy's in the linters, looks like that's not the case. 

Thanks for fixing!

StewartW (approved):

Issue #696, opened 2024-03-26
[Bug]: <Error during ADF deployment cfn_custom_resource>


giorgioGTelian:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

the errors arose during the initial phase of the deployment, initiated through the command `make deployment`. Despite ensuring all information was accurately filled in, including specifying the Organizational Unit (OU), we encountered an unexpected error.
The attached images indicate that this issue stems from a component called `cfn_custom_resource`, which seems to be the source of the complication.
you can vision the errors in the properties `Organization` and `DetermineFefaultBranchname`

### Expected Behavior

I expected the cloudFormation service to be successful without any error.

### Current Behavior

this is the current behaviour as of it now it gaves an error in the properties `Organization` and `DetermineFefaultBranchname`

![image (1)](https://github.com/awslabs/aws-deployment-framework/assets/7092023/20852c2e-593c-49ff-9450-735009a71db3)
![image](https://github.com/awslabs/aws-deployment-framework/assets/7092023/f1fa11cf-5fa8-4c48-b638-5586c91ae80b)

### Steps To Reproduce

1. git clone https://github.com/awslabs/aws-deployment-framework
2. cd aws-deployment-framework
3. git fetch --verbose --tags
4. git checkout ${version_tag_goes_here}
5. make update_makefile
6. make and make clean
7. make deploy putting the informations and creating the mandatory `deployment` OU in the AWS account

### ADF Version

the latest version

### Contributing a fix?

- [ ] Yes, I am working on a fix to resolve this issue

giorgioGTelian:
hello,
this is the report:  
```
*** Beginning of ADF Version Report ***
ADF Source version: v3.2.0
ADF SemanticVersion: 3.2.0

x86_64
Hardware platform: x86_64
Kernel name: Linux
Kernel release: 5.4.0-174-generic

OS Release:
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.6 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
[...]

PR #702, opened 2024-03-28
Install: Add checks to ensure installer dependencies are available

sbkok:
## Why?

When running the make / install process of ADF, we need to ensure it will not continue if some of the tools are not installed yet.

For example, in the past, if you did not install jq, it would still continue with the build process. Resulting in a broken template at deployment time.

As reported in #696, the Python version in the environment was set to v3.8, which resulted in a broken installation process too.

## What?

* Updated the documentation to indicate that `jq` needs to be available too.
* Added checks in the build process to exit and warn the user about missing dependencies.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Makefile versions
-MAKEFILE_VERSION := 2.0
+MAKEFILE_VERSION := 2.1
 UPDATE_VERSION := make/latest
+
+# This Makefile requires Python version 3.9 or later
+REQUIRED_PYTHON_MAJOR_VERSION := 3
+REQUIRED_PYTHON_MINOR_VERSION := 9
+PYTHON_EXECUTABLE := "python$(REQUIRED_PYTHON_MAJOR_VERSION)"
 
 # Repository versions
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -175,6 +175,48 @@ docs:
 	@echo "* $(CLR_BLUE)$(SRC_TAGGED_URL_BASE)/docs/user-guide.md$(CLR_END)"
 	@echo ""
 
+verify_tooling: .venv
+	@( \
+		. .venv/bin/activate; \
+		python3 --version &> /dev/null && \
+		( \
+			python3 -c "import sys; sys.version_info < (3,11) and sys.exit(1)" || \
```
javydekoning, line 183: I'm not a fan of where the minimum version is stored and compared (e.g. in the middle of the make file). Not sure what a better approach would be, but this may be hard to find for another maintainer.

sbkok: Good point, I will try to move it to the top somehow.
I'm trying to keep the Makefile one single file so it is easy to use with older versions of ADF as well.

sbkok: I've updated the Makefile to include the version number definition at the top as requested.

StewartW (approved):

javydekoning (approved):

PR #703, opened 2024-03-28
Fix clean bootstrap of the deployment account

sbkok:
## Why?

When you attempt to install ADF the first time, it will bootstrap the deployment account via the account bootstrap state machine. This, however, happens just before executing the bootstrap pipeline the first time. Resulting in missing parameters in the deployment account at the time the [regional stack](https://github.com/awslabs/aws-deployment-framework/blob/7a4359fcc441f3e62818b3c24569ddf069ec7a74/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/regional.yml#L14) is deployed.

This stack requires the `extensions/terraform/enabled` parameter (see [here](https://github.com/awslabs/aws-deployment-framework/blob/7a4359fcc441f3e62818b3c24569ddf069ec7a74/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/regional.yml#L14)). If that is not there yet, which is the case in the situation above, it will fail to bootstrap the account. Hereby the bootstrap state machine and pipeline both fail initially.

## What?

When the bootstrap state machine determines the event details, it should try to retrieve the parameter in the management account to see if Terraform is enabled or not. If this is not configured yet, it should default to False.

---

### DIFF
--- a/src/lambda_codebase/event.py
+++ b/src/lambda_codebase/event.py
@@ -86,12 +86,10 @@ class Event:
             1 if self.destination_ou_name == DEPLOYMENT_ACCOUNT_OU_NAME
             else 0
         )
-        try:
-            self.deployment_account_id = (
-                self.parameter_store.fetch_parameter('deployment_account_id')
-            )
-        except ParameterNotFoundError:
-            self.deployment_account_id = self.account_id
+        self.deployment_account_id = self._read_parameter(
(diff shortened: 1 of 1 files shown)


javydekoning (approved):

StewartW (approved):
Nice

PR #709, opened 2024-04-06
feat: BASH_SHFMT linter + linter fixes

javydekoning:
# Why?

For consistent formatting of bash scripts. 

*Issue #, if available:*

## What?

Added BASH_SHFMT

---


samples/sample-ec2-java-app-codedeploy/scripts/start.sh
```diff
@@ -2,4 +2,4 @@
 
 cd /home/ec2-user/server
 sudo /usr/bin/java -jar -Dserver.port=80 \
-  *.jar > /dev/null 2> /dev/null < /dev/null &
+  *.jar >/dev/null 2>/dev/null </dev/null &
```
sbkok, line 5: Might be me, but personally I like the space after the `>` sign.
It makes things more readable I think. What are your thoughts?
And can we change this setting?

A bit surprising that it does enforce spaces around `|`.

samples/sample-ec2-with-codedeploy/scripts/install-deps.sh
```diff
@@ -34,7 +34,7 @@ echo "<VirtualHost *:80>
   ProxyRequests Off
   ProxyPass / http://localhost:8080/
   ProxyPassReverse / http://localhost:8080/
-</VirtualHost>" >> sudo /etc/httpd/conf/httpd.conf
+</VirtualHost>" /etc/httpd/conf/httpd.conf >>sudo
```
sbkok, line 37: This would not work I think, would it?

sbkok (dismissed):

samples/sample-ec2-with-codedeploy/scripts/install-deps.sh
javydekoning: That does look odd... Let me move this to draft and investigate

javydekoning: I modified this as follows:

```sh
# Append VirtualHost configuration
echo "
<VirtualHost *:80>
  ProxyRequests Off
  ProxyPass / http://localhost:8080/
  ProxyPassReverse / http://localhost:8080/
</VirtualHost>
" | sudo tee -a /etc/httpd/conf/httpd.conf >/dev/null
```

samples/sample-ec2-java-app-codedeploy/scripts/start.sh
javydekoning: I agree with you. Adding `space_redirects = true` to editorconf to fix this.

sbkok (approved):
LGTM, thank you!

StewartW (approved):

PR #716, opened 2024-04-09
Fix sample expunge VPC, if-len, and process deployment maps

sbkok:
## Why?

Amazon CodeGuru suggested improvements to:

* Use paginators in the sample-expunge-vpc code, after reviewing, this sample required some more attention.
* Change `if len(...) > or ==` statements to match against the value. A list with elements is True, an empty list is False. Improving readability.
* Move boto3 client and resource creation to the initialization phase of the process deployment map Lambda function, to speed up its execution in consecutive runs.

---

### DIFF
--- a/samples/sample-expunge-vpc/src/lambda_vpc/lambda_function.py
+++ b/samples/sample-expunge-vpc/src/lambda_vpc/lambda_function.py
@@ -1,11 +1,12 @@
 # Copyright Amazon.com Inc. or its affiliates.
 # SPDX-License-Identifier: MIT-0
 
-from crhelper import CfnResource
-import logging
-import boto3
 from os import environ
+import logging
 import hashlib
+
+from crhelper import CfnResource
+import boto3
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


samples/sample-expunge-vpc/src/lambda_vpc/lambda_function.py
```diff
@@ -153,50 +177,56 @@ def get_regions(client):
...
+        if not default_vpc_found:
             logger.info("Creating default VPC in region")
             ec2_client.create_default_vpc()
-        else:
-            for vpc in vpcs['Vpcs']:
-                if not vpc['IsDefault']:
-                    logger.info("Creating default VPC in region")
-                    ec2_client.create_default_vpc()
 
         logger.info('~' * 72)
```
javydekoning, line 226: Can we get rid of this?

javydekoning (approved):

samples/sample-expunge-vpc/src/lambda_vpc/lambda_function.py
sbkok: When it iterates between accounts, the line is added to the logs to indicate that it switches from one account to another. So it is easier to trace. I looked into that line as well before, but kept it as I think it adds value.

StewartW (approved):

PR #718, opened 2024-04-24
Feat: Improve samples

sbkok:
## Why?

These changes aim to improve the overall quality, maintainability, and
usability of the ADF sample templates, while also providing better
documentation and aligning with the latest AWS best practices.

- To adhere to best practices and improve maintainability of the code base.
- To enhance security and observability by enabling VPC Flow Logs.
- To address code style issues and enforce consistent formatting.

## What?

- Specified explicit files to include in the CodeBuild output artifacts.
- Improvements to the VPC example:
      - Added support for enabling VPC Flow Logs to S3 or CloudWatch Logs.
      - Refactored resource ordering and added conditions for better readability.
      - Exported additional VPC CIDR range output for convenience.
- Fixed minor documentation issues in sample guides.
- Update README files with additional details, prerequisites, and deployment
   instructions for various samples.
- Upgrade the `sample-fargate-node-app` to use the AWS public container registry
   (public.ecr.aws) instead.
- Refactor the `sample-ec2-with-codedeploy` sample to use AWS Launch Templates,
   a newer and recommended approach instead of Launch Configurations.
- Update `sample-ec2-with-codedeploy` scripts to install newer versions of
   Amazon Linux 2023, Java, and other dependencies. Also fixed the scripts to be
   compatible to the recommended IMDSv2 authenticated APIs.
- Miscellaneous improvements and bug fixes across various sample templates.

---

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

PR #708, opened 2024-04-05
Adding retry logic for #655 and add tests for delete_default_vpc.py

javydekoning:
# Why?

Add retry logic with exponential backoff for #655 

*Issue #, if available:*

## What?

Description of changes:

- Added tests
- Added retry logic

PR #706, opened 2024-04-05
Moving CDK example app to latest CDK version

javydekoning:
# Why?

Move CDK sample app to CDKv2 for 4.0.0 release

## What?

Fixes #618 

---

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

PR #676, opened 2024-01-12
Upgrade CDK (v2.118), SAM (v1.107), Jinja2 (v3.1.3), and others to latest compatible version

sbkok:
## Why?

To use the latest versions that are compatible.

## What?

* Upgrade CDK (v2.118), SAM (v1.107), and others to latest compatible version
* Upgrading all remaining Lambda function to the ARM64 architecture was not possible.
   Unfortunately the `ruamel-yaml-clib==0.2.8(wheel)` dependency error popped up.
* Add the new required ImageId property for Cloud9 instances in the service
   catalog sample.
* Update to Python v3.12.
* Ensure Regular Expressions use raw strings where required.
* Remove Python based awscli v1, rely on default AWS CLI part of CodeBuild
   instead.
* Fix Terraform Enabled Parameter Store test
* Hide logs emitted by tests.
* Bumps [jinja2](https://github.com/pallets/jinja) from 3.1.2 to 3.1.3.
    - [Release notes](https://github.com/pallets/jinja/releases)
    - [Changelog](https://github.com/pallets/jinja/blob/main/CHANGES.rst)
    - [Commits](https://github.com/pallets/jinja/compare/3.1.2...3.1.3)

---

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

PR #732, opened 2024-05-20
Release v4.0.0

### LINKED ISSUE #20: Allow pipelines to target nested organisational units
Hi,

At the moment we can target a specific ou in the deployment_map.yml and roll out resources in all accounts within that ou. What would be great is if we can target accounts that are in a nested OU. Let me try to explain using some examples:

This would be our deployment map:
```yaml
 pipelines:
    type: cc-cloudformation # The pipeline_type you wish to use for this pipeline
    params:
      - SourceAccountId: 112233
      - NotificationEndpoint: me@moorspots.com
      - RestartExecutionOnUpdate: true
      - UpdateNestedOUs: true
    targets: 
      - path: /europe
        regions: eu-west-1
```

And this is our organisational structure:
```
- Root
   -- europe
        -> account0
        -- Development
            -> account1
            -> account2
        -- Production
            -> account3
```

At the moment the pipeline would only deploy stuff directly in the europe OU, in this case account0. Would it be possible to get adf to include the account1, account2 and account3 as well? 

Of course people might want to be able to tweak this behaviour so perhaps introducting a param like UpdateNestedOUs: true/false can help with this.

Cheers,
Thiezn

### LINKED ISSUE #355: S3 Role not working
I currently have this deploymentmap

```yaml
pipelines:
  - name: ben-test-pipeline
    tags: &pipeline_tags
      client: sandbox-ben
    default_providers:
      source:
        provider: codecommit
        properties:
          account_id: &sandbox_ben "111111111111"
          repository: adf-test-source
          branch: main
    targets:
      - path: *sandbox_ben
        provider: s3
        properties:
          bucket_name: demo-ben-202101
          object_key: output/aws-landing-zone-configuration.zip
          role: adf-custom-s3-role
```
That does however not generate the right pipeline, if I execute
```bash
aws codepipeline get-pipeline --name adf-pipeline-ben-test-pipeline --output yaml
```
I get
```yaml
# Removed the irrelevant sections
# ...
pipeline:
  name: adf-pipeline-ben-test-pipeline
    # ...
  roleArn: arn:aws:iam::222222222222:role/adf-codepipeline-role
  stages:
  - actions: # ...
    name: Source-111111111111
  - actions: # ...
    name: Build
  - actions:
    - actionTypeId:
        category: Deploy
        owner: AWS
        provider: S3
        version: '1'
      configuration:
        BucketName: demo-ben-202101
        Extract: 'false'
        ObjectKey: output/aws-landing-zone-configuration.zip
      inputArtifacts:
      - name: ben-test-pipeline-build
      name: AccountName-eu-west-1
      outputArtifacts: []
      region: eu-west-1
      roleArn: arn:aws:iam::111111111111:role/adf-cloudformation-role
      runOrder: 1
    name: deployment-stage-1
  version: 7
```

note that `roleArn` should be `arn:aws:iam::111111111111:role/adf-custom-s3-role`

I took a quick look at the code, and couldn't spot anything obvious that was wrong

### LINKED ISSUE #410: Support for multiple AWS Organizations
# Introduction

Typically, within your AWS Organization you have some setup to provide DTAP environments for your workloads. However testing the your implementation of the Landing Zone is not so easy to setup.

For simple tests like SCP you could use a test OU to first validate if the SCP work as intended. But not all cases could be tested that easily and you even still need to tweak the SCP when you move it. But other landing zone features like deploying CloudTrail in your master account and the bucket in your logging/archiving account then it becomes even harder.

For this reason we are setting up a test organization. This is replica of the production organization and also uses ADF. All repositories that deploy the landing zone resources deploy them from a development branch. While production deploys them from a master/main branch.

Some stacks that we created contained a hardcoded master, logging, etc account id. Because what are the chances that those will change right ;-). With the introduction of a test organization they do change.

## Goal

Collect information on what we need to change to ADF to support multiple organizations. And poll if there is a broader interest in this feature request. So, if you could use this 👍 this issue and leave your use-case, comments, suggestions in this thread.

### Maintaining duplicate config

You now need to use duplicate repos of the `aws-deployment-framework-bootstrap` and `aws-deployment-framework-pipelines`. Maintaining two repositories could lead to configuration drift over time.

#### Options

- Keep 2 different repositories, it gives the greatest form of flexibility.
- ?

### Code Repositories

When using CodeCommit the repositories live in the specific organization. But the two organizations are separated.

#### Options

- Create repositories in both organizations. Work only on 1 repository and sync a specific branch to the other organization for deployment. For example: if your working repo is in the testorg you could push master to the other repo when it's merged. But also the other way around. When you merge to development sync to the testorg when done testing merge to master.
  - We currently use a separate stack to do this. It's a Step Function that is triggered on a pull request being merged into development. When the event matches the requirements we will run a AWS Lambda that will checkout the code, assume a role in the other account. And do a `git push`.
- Allow cross organization access to the repository.

### Parameters

When you deal with organization agnostic templates that you deploy using ADF. You have to pass in for example the organization id or the account ids of the master, logging, security accounts.

#### Options

- Manually render the `<account-name>_<region>.json` from the `buildspec.yml`
- ADF needs to pass in the `OrganizationID` by default to all CodeBuild Projects, and `generate_params.py` will look for param files like: `o-myorgid.yml`.
- Use SSM Parameter. Currently, the `organization_id` parameter already exists so adding it to the `params/global.yml` file like this:
  ```yaml
  Parameters:
    OrganizationId: resolve:organization_id
    MasterAccountId: resolve:master_account_id
  ```
  Within ADF there is also a `master_account_id` parameter, so that is already available. But it would be nice if ADF manages a parameter for all AWS Account IDs. However, the account name and alias are different because it needs to be unique so we need a way to reference these accounts in a more generic way.
  By adding something like a unique goal to an account we could identify the accounts more easily:
  ```yaml
  - account_full_name: acme-testorg-logging
    organizational_unit_path: /my/path
    email: redacted@redacted.com
    alias: acme-testorg-logging
    goal: logging
  ```
  For example we could do something like `/adf/accounts/<goal>/account_id` so for your logging account that would be: `/adf/accounts/logging/account_id`. This allows you to use the following snippet:
  ```yaml
  Parameters:
    LoggingAccountId: resolve:/adf/accounts/logging/account_id
  ````
  
Open for any feedback!

### LINKED ISSUE #472: [ADF bootstrap] using IAM trust relations with internal AWS role IDs lead to issues
Hi dear ADF team,

Due to broken ADF pipeline creations we needed to recreate our `adf-global-base-deployment` stack within our deployment account.
This led to a recreation of the underlying `adf-codepipeline-role` Role.
For the deployments (e.q. Cloudformation) the trust relations to the `adf-cloudformation-role` is set to

```yaml
          - Effect: Allow
            Principal:
              AWS:
                - !Sub arn:aws:iam::${DeploymentAccountId}:role/adf-codepipeline-role
                - !Sub arn:aws:iam::${DeploymentAccountId}:role/adf-cloudformation-role
            Action:
              - sts:AssumeRole
```

which will be `AWS` internal replaced with the internal AWS role ID.
Due to this fact we run into several issues.

[@AndyEffa](https://github.com/AndyEfaa) had the idea to switch to a condition based account principal  so that such errors do not occur anymore.

Example:

```yaml
- Effect: Allow
            Action: 
              - 'sts:AssumeRole'
            Condition:
              StringEquals:
               'aws:PrincipalArn': 
                  - !Sub 'arn:aws:iam::${DeploymentAccountId}:role/adf-codepipeline-role'
                  - !Sub 'arn:aws:iam::${DeploymentAccountId}:role/adf-cloudformation-role'
            Principal:
              AWS: !Join
                - ''
                - - 'arn:'
                  - !Ref 'AWS::Partition'
                  - !Sub ':iam::${DeploymentAccountId}:root'
```

We want to throw in this idea to help other people avoid this issue.

### LINKED ISSUE #503: Cannot use BuildImages AMAZON_LINUX_2_4 or STANDARD_6_0 in deployment_maps
Hi,

In our organization we have started to upgrade our lambdas and other projects to nodejs16.x and we have a couple of ADF CodeBuild projects which use ubuntu "STANDARD_6_0" or even "AMAZON_LINUX_2_4". 

I noticed that our ADF master pipeline is using cdk@1.88 and the newer cdk@2.39.0 (for example) has in the class LinuxBuildImage: STANDARD_6_0 and AMAZON_LINUX_2_4. 
Reference: 
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_codebuild.LinuxBuildImage.html
https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-codebuild.LinuxBuildImage.html

At the moment if I need to update the BuildImage, I do it manually in the aws interface. 
I wonder, to do it via code as we normally do, is there a possibility to integrate this in the deployment_maps yml files ? Or do I need to upgrade ADF somehow ?

Here is also the error when trying to use AMAZON_LINUX_2_4 for example:
``` 

2022-08-26 07:56:02,180 \| INFO \| cdk_stacks.main \| Pipeline creation/update of my-FRONTEND-dev commenced \| (main.py:31)
--
334 | Traceback (most recent call last):
335 | File "adf-build/cdk/generate_pipeline_stacks.py", line 37, in <module>
336 | main()
337 | File "adf-build/cdk/generate_pipeline_stacks.py", line 33, in main
338 | PipelineStack(app, stack_input)
339 | File "/codebuild/output/src839325897/src/adf-build/jsii/_runtime.py", line 83, in __call__
340 | inst = super().__call__(*args, **kwargs)
341 | File "/codebuild/output/src839325897/src/adf-build/cdk/cdk_stacks/main.py", line 63, in __init__
342 | adf_codebuild.CodeBuild(
343 | File "/codebuild/output/src839325897/src/adf-build/jsii/_runtime.py", line 83, in __call__
344 | inst = super().__call__(*args, **kwargs)
345 | File "/codebuild/output/src839325897/src/adf-build/cdk/cdk_constructs/adf_codebuild.py", line 83, in __init__
346 | build_image=CodeBuild.determine_build_image(scope, target, map_params),
347 | File "/codebuild/output/src839325897/src/adf-build/cdk/cdk_constructs/adf_codebuild.py", line 186, in determine_build_image
348 | return getattr(
349 | AttributeError: type object 'LinuxBuildImage' has no attribute 'AMAZON_LINUX_2_4'
350 | Subprocess exited with error 1
351 |  
352 | [Container] 2022/08/26 07:56:02 Command did not exit successfully cdk synth --app adf-build/cdk/generate_pipeline_stacks.py 1> /dev/null exit status 1
353 | [Container] 2022/08/26 07:56:02 Phase complete: BUILD State: FAILED
```

Thank you,
Mircea

### LINKED ISSUE #594: [Bug]: Fresh installation of v3.2.0 fails to bootstrap the deployment account (upgrades work)

### LINKED ISSUE #597: [Bug]: AccountManagementStateMachine does not start when creating new account.

### LINKED ISSUE #599: [Bug]: Terraform script execution error
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Unable to execute Terraform pipeline due to an error in CodeBuild stage

### Expected Behavior

Module get_account.py should run with no error

### Current Behavior

During Terraform pipeline execution CodeBuild return the following error:

`Traceback (most recent call last):
  File "adf-build/helpers/terraform/get_accounts.py", line 190, in <module>
    main()
  File "adf-build/helpers/terraform/get_accounts.py", line 36, in main
    accounts_from_ous = get_accounts_from_ous()
  File "adf-build/helpers/terraform/get_accounts.py", line 95, in get_accounts_from_ous
    organizations = get_boto3_client(
  File "adf-build/helpers/terraform/get_accounts.py", line 145, in get_boto3_client
    role = sts.assume_role(
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/botocore/client.py", line 514, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/botocore/client.py", line 901, in _make_api_call
    request_dict = self._convert_to_request_dict(
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/botocore/client.py", line 962, in _convert_to_request_dict
    request_dict = self._serializer.serialize_to_request(
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/botocore/validate.py", line 381, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter RoleArn, value: ('arn:aws:sts::778473862620:role/OrganizationAccountAccessRole-readonly',), type: <class 'tuple'>, valid types: <class 'str'>`

### Steps To Reproduce

_No response_

### Possible Solution

Change the type of input parameter from tupla to string

`organizations = get_boto3_client(
        'organizations',
        f'arn:{PARTITION}:sts::{MANAGEMENT_ACCOUNT_ID}:role/{CROSS_ACCOUNT_ACCESS_ROLE}-readonly',
        'getaccountIDs',
    )`

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #608: [Bug]: Pipelines not triggered automatically on changes in source CodeCommit repository
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Although there is an Event Rule created for this in `adf-build/shared/cdk/cdk_constructs/adf_events.py`, the deployment is not triggered automatically since there is not a policy set in the default Event Bus (deployment account) that allows the source account to perform PutEvent actions on it.

### Expected Behavior

Pipelines to start automatically on source repositories changes.

### Current Behavior

Pipelines are not executed automatically

### Steps To Reproduce

1. Push some change to a codecommit repository in the source account
2. Check in the deployments account that the corresponding pipeline is not triggered automatically

### Possible Solution

Add the following policy to the Event Bus:

`{
  "Version": "2012-10-17",
  "Statement": [{
    "Sid": "allow_account_to_put_events",
    "Effect": "Allow",
    "Principal": {
      "AWS": "arn:aws:iam::${SOURCE_ACCOUNT}:root"
    },
    "Action": "events:PutEvents",
    "Resource": "arn:aws:events:eu-west-1:${DEPLOYMENTS_ACCOUNT}:event-bus/default"
  }]
}`

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #615: [Bug]: Upgrade to ADF 3.2.0 fails due to non-existing cross-account-role in child OUs
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

We are trying to upgrade our ADF installation to 3.2.0 but the enablecrossaccountaccess lambda fails for account in a child OU.

When reviewing the code, it looks like the role adf-update-cross-account-access-role exists in the adf-bootstrap/global.yml of the installation and all accounts using this yml seems to work.

However, we also have a subdir at adf-bootstrap/child-ou/global.yml which is not updated by the ADF upgrade. This file does not contain adf-update-cross-account-access-role which means that all accounts in this OU fails in the step function and eventually the upgrade times out and fails. 

Should we be adding the adf-update-cross-account-access-role manually for child OUs before upgrading? If so maybe this should be part of the installation/upgrade instructions?

When adding the role manually in the sub-OU global.yml, we fail instead on that the CloudFormationPolicyS3 is missing, should this be added also?

We are also a bit confused since we have inputted the AWSControlTowerExecution role as the crossaccount role, but this one is not used for this step for some reason?

Thanks a lot for your support!

BR
Gustav

### Expected Behavior

ADF Upgrade succeeds and uses the AWSControlTowerExecution role for crossaccount access

### Current Behavior

The ADF upgrade fails because the step function for cross-account-access times out with fails on accounts in child OU

### Steps To Reproduce

_No response_

### Possible Solution

_No response_

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #616: [Bug]: Permission issue when using Github as source
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

After upgrade to ADF 3.2.0 our existing aws-deployment-framework-base pipeline in the deployment account failed with `authorization failed for primary source and source version RANDSTR` after switching the source to Github. 

The reason for this is that the role used for the CodeBuild pipelines were changed in [this commit](https://github.com/awslabs/aws-deployment-framework/commit/c890a7cf2278672035f34efaf241ebd481f21216#diff-bdf00b4d802863247b08368c608a339058cf22cdeb3c75b427d153a4557e4946L546). The change removed the `codestar-connections` permissions needed to use Github as a source.

After adding the permissions back to the PipelineGenerationProvisionerCodeBuildRolePolicy we were able to successfully deploy using the codestar connection to our Github organization.

### Expected Behavior

To be able to use Github as a source for the pipeline aws-deployment-framework-base successfully.

### Current Behavior

Using Github as a source for aws-deployment-framework-base currently causes a authorization failure.

### Steps To Reproduce

_No response_

### Possible Solution

Add codestar permissions back to the PipelineGenerationProvisionerCodeBuildRolePolicy.

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #618: [Feat]: Update CDK Sample App to modern version
### Describe the feature

The sample code is still referencing node v8 and an old version of CDK

### Use Case

N/A

### Proposed Solution

Update https://github.com/awslabs/aws-deployment-framework/blob/f60bfaff93fb0d73b46bf7ff332f4c264cd3044d/samples/sample-cdk-app/package.json 

### Acknowledgements

- [X] I may be able to implement this feature request

### LINKED ISSUE #626: [Feat]: Do not hardcode DEFAULT_CODEBUILD_IMAGE
### Describe the feature

Today the `DEFAULT_CODEBUILD_IMAGE` is hardcoded here to an outdated Ubuntu 14.04 image:

https://github.com/awslabs/aws-deployment-framework/blob/a4de9abff63cf16fa83d55f7ae08829562fc4697/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py#L22

We have a few options to fix this.

1: Move the default to `adfconfig.yml` as suggested by @sbkok 
2: Default to the latest STANDARD image using this API: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/codebuild/client/list_curated_environment_images.html

Whatever option we take, we should also consider logging a warning in the adf build log that version pinning a build image in your pipeline is recommended.

Both options have the benefit of removing it from ADF upgrades.

### Use Case

N/A

### Proposed Solution

_No response_

### Acknowledgements

- [ ] I may be able to implement this feature request

### LINKED ISSUE #640: [Bug]: Long and similar pipelines names in same deployment map prevents pipeline from being created
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When pipelines with similar names that differ after the 30th character are defined in the same deployment map the ADFPipelineManagementStateMachine state machine is not triggered correctly.

Consider the following deployment map `mydeploymentmap.yml`

```yaml
pipelines:
  - name: kpi-dataplatform-qwerty-logistic
    default_providers:
      source:
        provider: codecommit
        properties:
          account_id: [...]
      build:
        provider: codebuild
        properties:
          image: "STANDARD_5_0"
          environment_variables:
            CONTAINS_TRANSFORM: False

  - name: kpi-dataplatform-qwerty-logistic-dev
    default_providers:
      source:
        provider: codecommit
        properties:
          repository: kpi-dataplatform-kupit-logistic
          branch: dev
          account_id: [...]
      build:
        provider: codebuild
        properties:
          image: "STANDARD_5_0"
          environment_variables:
            CONTAINS_TRANSFORM: False
```

### Expected Behavior

ADFPipelineManagementStateMachine gets executed twice and two CodePipeline pipelines get created.

### Current Behavior

ADFPipelineManagementStateMachine gets executed only once and only the first pipeline gets created

### Steps To Reproduce

Deploy the provided example deployment map

### Possible Solution

_No response_

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #642: [Bug]: terraform plan doesn't catch errors

### LINKED ISSUE #644: [Bug]: aws-deployment-framework-base pipeline fails
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Updating a deployment map and then running the aws-deployment-framework-base pipeline fails with 

```
[Container] 2023/07/18 14:22:14 Running command npm install aws-cdk@2.79 -g -y --quiet --no-progress

added 1 package in 2s

[Container] 2023/07/18 14:22:32 Running command aws s3 cp s3://$SHARED_MODULES_BUCKET/adf-build/ ./adf-build/ --recursive --quiet

[Container] 2023/07/18 14:22:50 Running command pip install -r adf-build/requirements.txt -r adf-build/helpers/requirements.txt -q -t ./adf-build
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [68 lines of output]
      /tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/config/setupcfg.py:293: _DeprecatedConfig: Deprecated config in `setup.cfg`
      !!
      
              ********************************************************************************
              The license_file parameter is deprecated, use license_files instead.
      
              By 2023-Oct-30, you need to update your project and remove deprecated calls
              or your builds will no longer be supported.
      
              See https://setuptools.pypa.io/en/latest/userguide/declarative_config.html for details.
              ********************************************************************************
      
      !!
        parsed = self.parsers.get(option_name, lambda x: x)(value)
      running egg_info
      writing lib3/PyYAML.egg-info/PKG-INFO
      writing dependency_links to lib3/PyYAML.egg-info/dependency_links.txt
      writing top-level names to lib3/PyYAML.egg-info/top_level.txt
      Traceback (most recent call last):
        File "/root/.pyenv/versions/3.11.2/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
          main()
        File "/root/.pyenv/versions/3.11.2/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/root/.pyenv/versions/3.11.2/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 341, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 323, in _get_build_requires
          self.run_setup()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/build_meta.py", line 338, in run_setup
          exec(code, locals())
        File "<string>", line 271, in <module>
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/__init__.py", line 107, in setup
          return distutils.core.setup(**attrs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 185, in setup
          return run_commands(dist)
                 ^^^^^^^^^^^^^^^^^^
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/core.py", line 201, in run_commands
          dist.run_commands()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 969, in run_commands
          self.run_command(cmd)
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/dist.py", line 1234, in run_command
          super().run_command(command)
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
          cmd_obj.run()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/command/egg_info.py", line 314, in run
          self.find_sources()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/command/egg_info.py", line 322, in find_sources
          mm.run()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/command/egg_info.py", line 551, in run
          self.add_defaults()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/command/egg_info.py", line 589, in add_defaults
          sdist.add_defaults(self)
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/command/sdist.py", line 104, in add_defaults
          super().add_defaults()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/command/sdist.py", line 251, in add_defaults
          self._add_defaults_ext()
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/command/sdist.py", line 336, in _add_defaults_ext
          self.filelist.extend(build_ext.get_source_files())
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "<string>", line 201, in get_source_files
        File "/tmp/pip-build-env-jayf5so3/overlay/lib/python3.11/site-packages/setuptools/_distutils/cmd.py", line 107, in __getattr__
          raise AttributeError(attr)
      AttributeError: cython_sources
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.

[notice] A new release of pip is available: 23.1.1 -> 23.2
[notice] To update, run: pip install --upgrade pip

[Container] 2023/07/18 14:23:14 Command did not exit successfully pip install -r adf-build/requirements.txt -r adf-build/helpers/requirements.txt -q -t ./adf-build exit status 1
[Container] 2023/07/18 14:23:14 Phase complete: INSTALL State: FAILED
[Container] 2023/07/18 14:23:14 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: pip install -r adf-build/requirements.txt -r adf-build/helpers/requirements.txt -q -t ./adf-build. Reason: exit status 1
```

### Expected Behavior

It's possible to run the pipeline and have it update the codepipelines according to the deployment maps

### Current Behavior

The pipeline updating the other pipelines fails.

### Steps To Reproduce

Update a deployment map
Push it to version control system
Trigger a build of the aws-deployment-framework-base pipeline

### Possible Solution

_No response_

### Additional Information/Context

_No response_

### ADF Version

3.2

### LINKED ISSUE #654: [Bug]: Description of EventBridge Rules for Completion Triggers shows variables placeholders
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When deployment maps contain completion triggers (e.g. to other pipelines) the description of the EventBridge Rules is not correctly parsed and it shows variables placeholders

### Expected Behavior

EventBridge Rule description looking like this
'Triggers MyDownStramPipeline on completion of MyUpStreamPipeline'

### Current Behavior

EventBridge Rule description is
'Triggers {pipeline} on completion of {params['pipeline']}'

### Steps To Reproduce

_No response_

### Possible Solution

_No response_

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #655: [Bug]: DeleteDefaultVPC fails with OptInRequired
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Osaka(ap-northeast-3) DeleteDefaultVPC fails in new account creation. This error only happens occasionally. I tried to reproduce it in our test - AWS Organizations but there DeleteDefaultVPC finished without any errors. 

### Expected Behavior

All default VPCs should be deleted after the account creation finishes.

### Current Behavior

AccountManagementStateMachine stops in 'Failed' status.

DeleteDefaultVPC Lambda has failed.
**Input**:
{
  "Payload": {
    "account_id": ".....",
    "region": "ap-northeast-3"
  }
}

**Output**:
{
  "errorMessage": "An error occurred (OptInRequired) when calling the DescribeVpcs operation: You are not subscribed to this service. Please go to http://aws.amazon.com to subscribe.",
  "errorType": "ClientError",
  "requestId": "485795bc-f736-422f-aeac-2ccd7df0f3d6",
  "stackTrace": [
    "  File \"/var/task/delete_default_vpc.py\", line 72, in lambda_handler\n    default_vpc_id = find_default_vpc(ec2_client)\n",
    "  File \"/var/task/delete_default_vpc.py\", line 28, in find_default_vpc\n    vpc_response = ec2_client.describe_vpcs()\n",
    "  File \"/var/task/botocore/client.py\", line 530, in _api_call\n    return self._make_api_call(operation_name, kwargs)\n",
    "  File \"/var/task/aws_xray_sdk/ext/botocore/patch.py\", line 38, in _xray_traced_botocore\n    return xray_recorder.record_subsegment(\n",
    "  File \"/var/task/aws_xray_sdk/core/recorder.py\", line 462, in record_subsegment\n    six.raise_from(exc, exc)\n",
    "  File \"<string>\", line 3, in raise_from\n",
    "  File \"/var/task/aws_xray_sdk/core/recorder.py\", line 457, in record_subsegment\n    return_value = wrapped(*args, **kwargs)\n",
    "  File \"/var/task/botocore/client.py\", line 960, in _make_api_call\n    raise error_class(parsed_response, operation_name)\n"
  ]
}

### Steps To Reproduce

_No response_

### Possible Solution

_No response_

### Additional Information/Context

_No response_

### ADF Version

3.2.0

### LINKED ISSUE #661: [Feat]: Terraform variables per region
### Describe the feature

For terraform pipelines we need a way to define different variables for different regions. This is crucial in scenarios where the same pipeline must run in different regions for a given account.

### Use Case

A pipeline that creates VPCs for an account needs as input the CIDR blocks to allocate, that most likely differ from region to region for an account.
With the current approach, you need to create a rather unnecessary complex variable map where all CIDR blocks for all regions are stored and then pick up the one needed in the current target region.
It can be much easier to separate variables per region.

### Proposed Solution

#662 

### Acknowledgements

- [X] I may be able to implement this feature request

### LINKED ISSUE #664: [Bug]: Account Based Terraform variables do not work

### LINKED ISSUE #698: [Chore]: Upgrade GitHub CodeQL Action v2 to v3
## Why?

As can be seen at the following [GitHub MegaLinter Job](https://github.com/awslabs/aws-deployment-framework/actions/runs/8461369655/job/23181032522?pr=697): when it attempts to upload the MegaLinter scan results to the GitHub Security tab, it states that CodeQL Action v2 will be deprecated on December 5th, 2024.

## What?

It seems we need to upgrade to v3: https://github.blog/changelog/2024-01-12-code-scanning-deprecation-of-codeql-action-v2/

sbkok:
## v4.0.0

This is a security-focused release of the AWS Deployment Framework (ADF) that
aims to restrict the default access required and provided by ADF via the
least-privilege principle.

__Key security enhancements include:__

- Applying IAM best practices by restricting excessive permissions granted to
  IAM roles and policies used by ADF.
- Leveraging new IAM features to further limit access privileges granted by
  default, reducing the potential attack surface.
- Where privileged access is required for specific ADF use cases, the scope and
  duration of elevated privileges have been minimized to limit the associated
  risks.

By implementing these security improvements, ADF now follows the principle of
least privilege, reducing the risk of unauthorized access or
privilege-escalation attacks.

Please make sure to go through the list of changes breaking changes carefully.

As with every release, it is strongly recommended to thoroughly review and test
this version of ADF in a non-production environment first.

#### Security: Confused Deputy Problem

Addressed the [Confused Deputy problem](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html)
in IAM roles created by ADF to use by the AWS Services. Where supported, the
roles are restricted to specific resources via an `aws:SourceArn` condition.
If you were using the ADF roles for other resources or use cases not covered
by ADF, you might need to patch the Assume Role policies accordingly.

#### Security: Cross-Account Access Role and the new Jump Role

ADF relies on the privileged Cross-Account Access Role to bootstrap accounts.
In the past, ADF used this role for every update and deployment of the
bootstrap stacks, as well as account management features.

With the release of v4.0, a jump role is introduced to lock-down the usage of
the privileged cross-account access role. Part of the bootstrap stack, the
`adf-bootstrap-update-deployment-role` is created. This role grants access to
perform restricted updates that are frequently performed via the
`aws-deployment-framework-bootstrap` pipeline. By default, the jump role is
granted access to assume into this update deployment role.

A dedicated jump role manager is responsible to grant the jump role access to
the cross-account access role for AWS accounts where ADF requires access and
the `adf-bootstrap-update-deployment-role` is not available yet.
For example, accounts that are newly created only have the cross-account access
role to assume into. Same holds for ADF managed accounts that are not updated
to the new v4.0 bootstrap stack yet.

During the installation/update of ADF, a new parameter enables you to grant
the jump role temporary access to the cross-account access role as an
privileged escalation path.
This parameter is called `GrantOrgWidePrivilegedBootstrapAccessUntil`.
By setting this to a date/time in the future you will grant access to the
cross-account access role until that date/time. This would be required if you
modify ADF itself or the bootstrap stack templates. Changing permissions like
the `adf-cloudformation-deployment-role` is possible without relying on the
cross-account access role. For most changes deployed via the bootstrap pipeline
it does not require elevated privileged access to update.

With the above changes, the `aws-deployment-framework-bootstrap` CodeBuild
project no longer has unrestricted access to the privileged cross-account role.
Starting from version 4.0, access to assume the privileged cross-account access
role is restricted and must be obtained through the Jump Role as described
above.

#### Security: Restricted account management access

Account Management is able to access non-protected organization units.
Prior to ADF v4.0, the account management process used the privileged
cross-account assess role to operate. Hence it could move an account or update
the properties of an account that is located in a protected organization unit
too. With the release of v4.0, it is only able to move or manage accounts if
they are accessible via the Jump Role. The Jump Role is restricted to
non-protected organization units only.

This enhances the security of ADF, as defining a organization unit as protected
will block access to that via the Jump Role accordingly.

#### Security: Restricted bootstrapping of management account

The `adf-global-base-adf-build` stack in the management account was initially
deployed to facilitate bootstrap access to the management account.
It accomplished this by creating a cross-account access role with limited
permissions in the management account ahead of the bootstrapping process.

ADF created this role as it is not provisioned by AWS Organizations or
AWS Control Tower in the management account itself. However, ADF required some
level of access to deploy the necessary bootstrap stacks when needed.

It is important to note that deploying this role and bootstrapping the
management account introduces a potential risk. A pipeline created via a
deployment map could target the management account and create resources within
it, which may have unintended consequences.

To mitigate the potential risk, it is recommended to implement strict
least-privilege policies and apply permission boundaries to protect
the management account.
Additionally, thoroughly reviewing all deployment map changes is crucial to
ensure no unintended access is granted to the management account.

With the release of ADF v4.0, the `adf-global-base-adf-build` stack is removed
and its resources are moved to the main ADF CloudFormation template.
These resources will only get deployed if the new
`AllowBootstrappingOfManagementAccount` parameter is set to `Yes`. By default
it will not allow bootstrapping of the management account.

#### Security: Restricted bootstrapping of deployment account

Considering the sensitive workloads that run in the deployment account, it is
important to limit the permissions granted for pipelines to deploy to the
deployment account itself. You should consider the deployment account a
production account.

It is recommended to apply the least-privilege principle and only allow
pipelines to deploy resources that are required in the deployment account.

Follow these steps after the changes introduced by the ADF v4.0 release are
applied in the main branch of the `aws-deployment-framework-bootstrap`
repository.

Please take this moment to review the following:

* Navigate to the `adf-boostrap/deployment` folder in that repository.
* Check if it contains a `global-iam.yml` file:

    * If it does __not__ contain a `global-iam.yml` file yet, please ensure you
      copy the `example-global-iam.yml` file in that directory.
    * If it does, please compare it against the `example-global-iam.yml` file
      in that directory.

* Apply the least-privilege principle on the permissions you grant in the
  deployment account.

#### Security: Shared Modules Bucket

ADF uses the Shared Modules Bucket as hosted in the management account in the
main deployment region to share artifacts from the
`aws-deployment-framework-bootstrap` repository.

The breaking change enforces all objects to be owned by the bucket owner from
v4.0 onward.

#### Security: ADF Role policy restrictions

With the v4.0 release, all ADF roles and policies were reviewed, applying
the latest best-practices and granting access to ADF resources only where
required. This review also includes the roles that were used by the pipelines
generated by ADF.

Please be aware of the changes made to the following roles:

##### adf-codecommit-role

The `adf-codecommit-role` no longer grants read/write access to all buckets.
It only grants access to the buckets created and managed by ADF where it
needed to. Please grant access accordingly if you use custom S3 buckets or need
to copy from an S3 bucket in an ADF-generated pipeline.

##### adf-codebuild-role

The `adf-codebuild-role` can only be used by CodeBuild projects in the main
deployment region. ADF did not allow running CodeBuild projects in other
regions before. But in case you manually configured the role in a project
in a different region it will fail to launch.

The `adf-codebuild-role` is no longer allowed to assume any IAM Role in the
target accounts if those roles would grant access in the Assume Role
Policy Document.

The `adf-codebuild-role` is restricted to assume only the
`adf-readonly-automation-role` roles in the target accounts.
And, in the case that the Terraform ADF Extension is enabled, it is allowed to
assume the `adf-terraform-role` too.

It is therefore not allowed to assume the `adf-cloudformation-deployment-role`
any longer. If you were deploying with `cdk deploy` into target accounts from an
ADF pipeline you will need to specifically grant the `adf-codebuild-role`
access to assume the `adf-cloudformation-deployment-role`. However, we strongly
recommend you synthesize the templates instead and let AWS CloudFormation do
the deployment for you.

For Terraform support, CodeBuild was granted access to the `adf-tflocktable`
table in release v3.2.0. This access is restricted to only grant read/write
access to that table if the Terraform extension is enabled.
Please bear in mind that if you enable Terraform access the first time, you
will need to use the `GrantOrgWidePrivilegedBootstrapAccessUntil` parameter
if ADF v4.0 bootstrapped to accounts before. As this operation requires
privileged access.

The `adf-codebuild-role` is allowed to assume into the
`adf-terraform-role` if the Terraform extension is enabled.
As written in the docs, the `adf-terraform-role` is configured
in the `global-iam.yml` file. This role is commented out by default.
When you define this role, it is important to make sure to grant it
[least-privilege access](https://docs.aws.amazon.com/IAM/latest/UserGuide/best-practices.html#grant-least-privilege) only.

##### adf-cloudformation-role

The `adf-cloudformation-role` is no longer assumable by CloudFormation.
This role is used by CodePipeline to orchestrate various deployment actions
across accounts. For example, CodeDeploy, S3, and obviously the CloudFormation
actions.

For CloudFormation, it would instruct the service to use the CloudFormation
Deployment role for the actual deployment.
The CloudFormation deployment role is the role that is assumed by the
CloudFormation service. This change should not impact you, unless you
use this role in relation with CloudFormation that is not managed by ADF.

With v4.0, the `adf-cloudformation-role` is only allowed to pass the
CloudFormation Deployment role to CloudFormation and no other roles to other
services.

If you were/want to make use of a custom CloudFormation deployment role for
specific pipelines, you need to make sure that the `adf-cloudformation-role` is
allowed to perform an `iam:PassRole` action with the given role.
It is recommended to limit this to be passed to the CloudFormation service
only. You can find an example of this in the
`adf-bootstrap/deployment/global.yml` file where it allows the
CloudFormation role to perform `iam:PassRole` with the
`adf-cloudformation-deployment-role`. When required, please grant this access
in the `adf-bootstrap/deployment/global-iam.yml` file in the
`aws-deployment-framework-bootstrap` repository.

Additionally, the `adf-cloudformation-role` is not allowed to access S3 buckets
except the ADF buckets it needs to transfer pipeline assets to CloudFormation.

##### adf-codepipeline-role

The `adf-codepipeline-role` is no longer assumable by CloudFormation,
CodeDeploy, and S3. The role itself was not passed to any of these services by
ADF.

If you relied on the permissions that were removed, feel free to extend the
role permissions via the `global-iam.yml` stack.

#### Security: Restricted access to ADF-managed S3 buckets only

With v4.0, access is restricted to ADF-managed S3 buckets only.
If a pipeline used the S3 source or deployment provider, it will require
the required access to those buckets. Please add the required access to the
`global-iam.yml` bootstrap stack in the OU where it is hosted.

Grant read access to the `adf-codecommit-role` for S3 source buckets.
Grant write access to the `adf-cloudformation-role` for S3 buckets an ADF
pipeline deploys to.

#### Security: Bootstrap stack no longer named after organization unit

The global and regional bootstrap stacks are renamed to
`adf-global-base-bootstrap` and `adf-regional-base-bootstrap` respectively.

In prior releases of ADF, the name ended with the organization unit name.
As a result, an account could not move from one organization unit to
another without first removing the bootstrap stacks. Additionally, it made
writing IAM policies and SCPs harder in a least-privilege way.

When ADF v4.0 is installed, the legacy stacks will get removed by the
`aws-deployment-framework-bootstrap` pipeline automatically. Shortly after
removal, it will deploy the new bootstrap stacks.

With v4.0, accounts can move from one organization unit to another,
without requiring the removal of the ADF bootstrap stacks.

#### Security: KMS Encryption required on Deployment Account Pipeline Buckets

The deployment account pipeline buckets only accepts KMS Encrypted objects from
v4.0 onward. Ensuring that all objects are encrypted with the same KMS Key.

Before, some objects used KMS encryption while others did not. The bucket
policy now requires all objects to be encrypted via the KMS key. All ADF
components have been adjusted to upload with this key. If, however, you copy
files from systems that are not managed by ADF, you will need to adjust these
to encrypt the objects with the KMS key as well.

#### Security: TLS Encryption required on all ADF-managed buckets

S3 Buckets created by ADF will require TLS 1.2 or later. All actions that occur
on these buckets with older TLS versions will be denied via the bucket policies
that these buckets received.

#### New installer

The dependencies that are bundled by the move to the AWS Cloud Development Kit
(CDK) v2 increased the deployment size of ADF.
Unfortunately it increased the deployment size beyond the limit that is
supported by the Serverless Application Repository (SAR).

Hence a new installation mechanism is required.

Please read the [installation instructions](https://github.com/awslabs/aws-deployment-framework/blob/master/docs/installation-guide.md) carefully.

In case you are upgrading an existing installation of ADF, please consider
following the [upgrade steps as defined in the admin guide](https://github.com/awslabs/aws-deployment-framework/blob/master/docs/admin-guide.md#updating-between-versions).

#### CDK v2

ADF v4.0 is built on the AWS Cloud Development Kit (CDK) v2. Which is an
upgrade to CDK v1 that ADF relied on before.

For most end-users, this change would not have an immediate impact.
If, however, you made customizations to ADF it might require you to upgrade
these customizations to CDK v2 as well.

#### CodeBuild default image

As written in the [CodeBuild provider docs](./docs/providers-guide.md#properties-3), it is a best-practice to define
the exact CodeBuild container image you would like to use for each pipeline.

However, in case you rely on the default, in prior ADF releases it would
default to `UBUNTU_14_04_PYTHON_3_7_1`. This container image is no longer
supported. With ADF v4.0, the new default is `STANDARD_7_0`.
Also referred to as: `aws/codebuild/standard:7.0`.

#### ADF Renaming of Roles

ADF v4.0 changes most of the roles that it relies on. The reason for this
change is to make it easier to secure ADF with Service Control Policies and
IAM permission boundaries. Where applicable, the roles received a new prefix.
This makes it easier to identify what part of ADF relies on those roles and
whom should have access to assume the role or modify it.

| Previous prefix  | Previous name                                                       | New prefix                 | New name                                                      |
|------------------|---------------------------------------------------------------------|----------------------------|---------------------------------------------------------------|
| /                | ${CrossAccountAccessRoleName}-readonly                              | /adf/organizations/        | adf-organizations-readonly                                    |
| /                | adf-update-cross-account-access-role                                | /adf/bootstrap/            | adf-update-cross-account-access                               |
| /adf-automation/ | adf-create-repository-role                                          | /adf/pipeline-management/  | adf-pipeline-management-create-repository                     |
| /adf-automation/ | adf-pipeline-provisioner-generate-inputs                            | /adf/pipeline-management/  | adf-pipeline-management-generate-inputs                       |
| /adf-automation/ | adf-pipeline-create-update-rule                                     | /adf/pipeline-management/  | adf-pipeline-management-create-update-rule                    |
| /                | adf-event-rule-${AWS::AccountId}-${DeploymentAccountId}-EventRole-* | /adf/cross-account-events/ | adf-cc-event-from-${AWS::AccountId}-to-${DeploymentAccountId} |
|------------------|---------------------------------------------------------------------|----------------------------|---------------------------------------------------------------|

#### ADF Renaming of Resources

| Type         | Previous name                                 | New name                                               |
|--------------|-----------------------------------------------|--------------------------------------------------------|
| StateMachine | EnableCrossAccountAccess                      | adf-bootstrap-enable-cross-account                     |
| StateMachine | ADFPipelineManagementStateMachine             | adf-pipeline-management                                |
| StateMachine | PipelineDeletionStateMachine-*                | adf-pipeline-management-delete-outdated                |
| Lambda       | DeploymentMapProcessorFunction                | adf-pipeline-management-deployment-map-processor       |
| Lambda       | ADFPipelineCreateOrUpdateRuleFunction         | adf-pipeline-management-create-update-rule             |
| Lambda       | ADFPipelineCreateRepositoryFunction           | adf-pipeline-management-create-repository              |
| Lambda       | ADFPipelineGenerateInputsFunction             | adf-pipeline-management-generate-pipeline-inputs       |
| Lambda       | ADFPipelineStoreDefinitionFunction            | adf-pipeline-management-store-pipeline-definition      |
| Lambda       | ADFPipelineIdentifyOutOfDatePipelinesFunction | adf-pipeline-management-identify-out-of-date-pipelines |
|--------------|-----------------------------------------------|--------------------------------------------------------|

#### ADF Parameters in AWS Systems Manager Parameter Store

Some of the parameters stored by ADF in AWS Systems Manager Parameter Store
were located at the root of the Parameter Store. This made it hard to maintain
and restrict access to the limited set of ADF specific parameters.

With ADF v4.0, the parameters used by ADF are located under the `/adf/` prefix.
For example, `/adf/deployment_account_id`.

The `global-iam.yml` bootstrap stack templates get copied from their
`example-global-iam.yml` counterparts. When this was copied in v3.2.0, the
default path for the `deployment_account_id` parameter should be updated to
`/adf/deployment_account_id`. Please apply this new default value to the
CloudFormation templates accordingly. If you forget to do this, the stack
deployment of the `adf-global-base-iam` stack might fail with a failure stating
that it does not have permission to fetch the `deployment_account_id`
parameter.

The error you run into if the parameter path is not updated:

> An error occurred (ValidationError) when calling the CreateChangeSet
> operation: User:
> arn:aws:sts::111111111111:assumed-role/${CrossAccountAccessRoleName}/base_update
> is not authorized to perform: ssm:GetParameters on resource:
> arn:aws:ssm:${deployment_region}:111111111111:parameter/deployment_account_id
> because no identity-based policy allows the ssm:GetParameters action
> (Service: AWSSimpleSystemsManagement; Status Code: 400;
> Error Code: AccessDeniedException; Request ID: xxx).

If an application or customization to ADF relies on one of these parameters
they will need to be updated to include this prefix. Unless the application
code relies on ADF's ParameterStore class, in that case it will automatically
prefix the `/adf/` to all parameters read or written.

With the changes in the IAM policies, ADF's access is restricted to the `/adf/`
prefix. This, unfortunately implies that old parameters are not deleted when
you update your installation of ADF. There is no cost associated to these
parameters, so you can leave them as is.
Feel free to delete the old parameters.

The parameters that are managed by ADF that got their path changed are:

For the __management account__, in the __AWS Organizations region__
(`us-east-1`, or `us-gov-west-1`):

| Old Parameter Path           | New Parameter Path               |
|------------------------------|----------------------------------|
| `/adf_log_level`             | `/adf/adf_log_level`             |
| `/adf_version`               | `/adf/adf_version`               |
| `/bucket_name`               | `/adf/bucket_name`               |
| `/confit`                    | `/adf/config`                    |
| `/cross_account_access_role` | `/adf/cross_account_access_role` |
| `/deployment_account_id`     | `/adf/deployment_account_id`     |
| `/deployment_account_region` | `/adf/deployment_account_region` |
| `/kms_arn`                   | `/adf/kms_arn`                   |
| `/notification_channel`      | `/adf/notification_channel`      |
| `/organization_id`           | `/adf/organization_id`           |
| `/protected`                 | `/adf/protected`                 |
| `/scp`                       | `/adf/scp`                       |
| `/shared_modules_bucket`     | `/adf/shared_modules_bucket`     |
| `/tagging-policy`            | `/adf/tagging_policy`            |
| `/target_regions`            | `/adf/target_regions`            |

For the __management account__, in __other ADF regions__:

| Old Parameter Path           | New Parameter Path               |
|------------------------------|----------------------------------|
| `/adf_version`               | `/adf/adf_version`               |
| `/bucket_name`               | `/adf/bucket_name`               |
| `/cross_account_access_role` | `/adf/cross_account_access_role` |
| `/deployment_account_id`     | `/adf/deployment_account_id`     |
| `/kms_arn`                   | `/adf/kms_arn`                   |

For the __deployment account__, in __the deployment region__:

| Old Parameter Path           | New Parameter Path                  |
|------------------------------|-------------------------------------|
| `/adf_log_level`             | `/adf/adf_log_level`                |
| `/adf_version`               | `/adf/adf_version`                  |
| `/auto_create_repositories`  | `/adf/scm/auto_create_repositories` |
| `/cross_account_access_role` | `/adf/cross_account_access_role`    |
| `/default_scm_branch`        | `/adf/scm//default_scm_branch`      |
| `/deployment_account_bucket` | `/adf/shared_modules_bucket`        |
| `/master_account_id`         | `/adf/management_account_id`        |
| `/notification_endpoint`     | `/adf/notification_endpoint`        |
| `/notification_type`         | `/adf/notification_type`            |
| `/organization_id`           | `/adf/organization_id`              |

For the __deployment account__, in __other ADF regions__:

| Old Parameter Path           | New Parameter Path               |
|------------------------------|----------------------------------|
| `/adf_log_level`             | `/adf/adf_log_level`             |
| `/adf_version`               | `/adf/adf_version`               |
| `/cross_account_access_role` | `/adf/cross_account_access_role` |
| `/deployment_account_bucket` | `/adf/shared_modules_bucket`     |
| `/master_account_id`         | `/adf/management_account_id`     |
| `/notification_endpoint`     | `/adf/notification_endpoint`     |
| `/notification_type`         | `/adf/notification_type`         |
| `/organization_id`           | `/adf/organization_id`           |

For a __target account__, in __each ADF region__:

| Old Parameter Path       | New Parameter Path           |
|--------------------------|------------------------------|
| `/bucket_name`           | `/adf/bucket_name`           |
| `/deployment_account_id` | `/adf/deployment_account_id` |
| `/kms_arn`               | `/adf/kms_arn`               |

#### AWS CodeStar Connections OAuth Token support dropped

ADF v4.0 discontinued the support for the OAuth Token stored in
SSM Parameter Store. As this method is not advised to be used by CodePipeline,
and might leave the OAuth Token accessible to other users of the deployment
account. As this is not a security best practice, ADF v4.0 no longer supports
it.

To upgrade, please read the [Administrator Guide on Using AWS CodeConnections
for Bitbucket, GitHub, or GitLab](./docs/admin-guide.md#using-aws-codeconnections-for-bitbucket-github-github-enterprise-or-gitlab).

#### AWS CodeStar Connections changed to AWS CodeConnections

The AWS CodeStar Connection service [changed its name to AWS
CodeConnections](https://docs.aws.amazon.com/dtconsole/latest/userguide/rename.html).

If you configured a CodeStar Connection before, you can continue to use that.
You do not need to update the CodeStar policy as defined in the
`aws-deployment-framework-bootstrap/adf-bootstrap/deployment/global-iam.yml`
stack.

However, please update the pipeline definitions in your deployment map files.
The changes you need to make are renaming the source
provider from `codestar` to `codeconnections`.
Also update the `codestar_connection_path` source property to
`codeconnections_param_path`.

Both of these changes can be seen in the following example:

```yaml
pipelines:
  - name: sample-vpc
    default_providers:
      source:
        # provider: codestar
        provider: codeconnections
        properties:
          # codestar_connection_path: /adf/my_connection_arn_param
          codeconnections_param_path: /adf/my_connection_arn_param
```

If you are upgrading from the GitHub OAuth token or otherwise require a new
source code connection, please proceed with the AWS CodeConnections
configuration as defined in the
[Admin Guide - Using AWS CodeConnections for Bitbucket, GitHub, or
GitLab](./docs/admin-guide.md#using-aws-codeconnections-for-bitbucket-github-or-gitlab).

### Features

- Update CDK from v1 to v2 (#619), by @pergardebrink, resolves #503, #614, and #617.
- Account Management State Machine will now opt-in to target regions when creating an account (#604) by @StewartW.
- Add support for nested organization unit targets (#538) by @StewartW, resolves #20.
- Enable single ADF bootstrap and pipeline repositories to multi-AWS Organization setup, resolves #410:
    - Introduce the org-stage (#636) by @AndyEfaa.
    - Add support to allow empty targets in deployment maps (#634) by @AndyEfaa.
    - Add support to define the "default-scm-codecommit-account-id" in adfconfig.yml, no value in either falls back to deployment account id (#633) by @AndyEfaa.
    - Add multi AWS Organization support to adfconfig.yml (#668) by @alexevansigg.
    - Add multi AWS Organization support to generate_params.py (#672) by @AndyEfaa.
- Terraform: add support for distinct variable files per region per account in Terraform pipelines (#662) by @igordust, resolves #661.
- CodeBuild environment agnostic custom images references, allowing to specify the repository name or ARN of the ECR repository to use (#623) by @abhi1094.
- Add kms_encryption_key_arn and cache_control parameters to S3 deploy provider (#669) by @alFReD-NSH.
- Allow inter-ou move of accounts (#712) by @sbkok.

### Fixes

- Fix Terraform terrascan failure due to incorrect curl call (#607), by @lasv-az.
- Fix custom pipeline type configuration not loaded (#612), by @lydialim.
- Fix Terraform module execution error (#600), by @stemons, resolves #599 and #602.
- Fix resource untagging permissions (#635) by @sbkok.
- Fix GitHub Pipeline secret token usage (#645) by @sbkok.
- Fix Terraform error masking by tee (#643) by @igordust, resolves #642.
- Fix create repository bug when in rollback complete state (#648) by @alexevansigg.
- Fix cleanup of parameters upon pipeline retirement (#652) by @sbkok.
- Fix wave calculation for non-default CloudFormation actions and multi-region deployments (#624 and #651), by @alexevansigg.
- Fix ChatBot channel ref + add notification management permissions (#650) by @sbkok.
- Improve docs and add CodeStar Connection policy (#649) by @sbkok.
- Fix Terraform account variables were not copied correctly (#665) by @donnyDonowitz, resolves #664.
- Fix pipeline management state machine error handling (#683) by @sbkok.
- Fix target schema for tags (#667) by @AndyEfaa.
- Fix avoid overwriting truncated pipeline definitions with pipelines that share the same start (#653) by @AndyEfaa.
- Fix updating old global-iam stacks in the deployment account (#711) by @sbkok.
- Remove default org-stage reference to dev (#717) by @alexevansigg.
- Fix racing condition on first-usage of ADF pipelines leading to an auth error (#732) by @sbkok.
- Fix support for custom S3 deployment roles (#732) by @sbkok, resolves #355.
- Fix pipeline completion trigger description (#734) by @sbkok, resolves #654.

### Improvements

- Sanitizing account names before using them in SFn Invocation (#598) by @StewartW, resolves #597.
- Improve Terraform documentation sample (#605), by @lasv-az.
- Fix CodeDeploy sample to work in gov-cloud (#609), by @sbkok.
- Fix documentation error on CodeBuild custom image (#622), by @abhi1094.
- Speedup bootstrap pipeline by removing unused SAM Build (#613), by @AlexMackechnie.
- Upgrade CDK (v2.88), SAM (v1.93), and others to latest compatible version (#647) by @sbkok, resolves #644.
- Update pip before installing dependencies (#606) by @lasv-az.
- Fix: Adding hash to pipelines processing step function execution names to prevent collisions (#641) by @avolip, resolves #640.
- Modify trust relations for roles to ease redeployment of roles (#526) by @AndreasAugustin, resolves #472.
- Limit adf-state-machine-role to what is needed (#657) by @alFReD-NSH.
- Upload SCP policies with spaces removed (#656) by @alFReD-NSH.
- Move from ACL enforced bucket ownership to Ownership Controls + MegaLinter prettier fix (#666) by @sbkok.
- Upgrade CDK (v2.119), SAM (v1.107), Jinja2 (v3.1.3), and others to latest compatible version (#676) by @sbkok.
- Fix initial value type of allow-empty-targets (#678) by @sbkok.
- Fix Shared ADF Lambda Layer builds and add move to ARM-64 Lambdas (#680) by @sbkok.
- Add /adf params prefix and other SSM Parameter improvements (#695) by @sbkok, resolves #594 and #659.
- Fix pipeline support for CodeBuild containers with Python < v3.10 (#705) by @sbkok.
- Update CDK v2.136, SAM CLI 1.114, and others (#715) by @sbkok.
- AWS CodeStar Connections name change to CodeConnections (#714) by @sbkok, resolves #616.
- Adding retry logic for #655 and add tests for delete_default_vpc.py (#708) by @javydekoning, resolves #655.
- Fix allow-empty-targets to match config boolean style (#725) by @sbkok.
- Require previously optional CodeBuild image property in build/deploy from v4 onward (#731) by @sbkok, resolves #626 and #601.
- YAML files are interpreted via `YAML.safe_load` instead of `YAML.load` (#732) by @sbkok.
- Hardened all urlopen calls by checking the protocol (#732) by @sbkok.
- Added check to ensure the CloudFormation deployment account id matches with the `/adf/deployment_account_id` if that exists (#732) by @sbkok.
- Add automatic creation of the `/adf/deployment_account_id` and `/adf/management_account_id` if that does not exist (#732) by @sbkok.
- Separate delete outdated state machine from pipeline creation state machines (#732) by @sbkok.
- Review and restrict access provided by ADF managed IAM roles and permissions (#732) by @sbkok, resolves #608 and #390.
- Add automatic clean-up of legacy bootstrap stacks, auto recreate if required (#732) by @sbkok.

#### Installation improvements

With the addition of CDK v2 support. The dependencies that go with it,
unfortunately increased the deployment size beyond the limit that is supported
by the Serverless Application Repository. Hence the SAR installer is replaced
by a new installation process.
Please read the [Installation Guide](https://github.com/awslabs/aws-deployment-framework/blob/make/latest/docs/installation-guide.md) how to install ADF.
In case you are upgrading, please follow [the admin guide on updating ADF](https://github.com/awslabs/aws-deployment-framework/blob/make/latest/docs/admin-guide.md#updating-between-versions) instead.

- New installation process (#677) by @sbkok.
- Auto generate unique branch names on new version deployments (#682) by @sbkok.
- Ensure tox fails at first pytest failure (#686) by @sbkok.
- Install: Add checks to ensure installer dependencies are available (#702) by @sbkok.
- Install: Add version checks and pre-deploy warnings (#726) by @sbkok.
- Install: Add uncommitted changes check (#733) by @sbkok.

#### Documentation, ADF GitHub, and code only improvements

- Fixing broken Travis link and build badge (#625), by @javydekoning.
- Temporarily disabled cfn-lint after for #619 (#630), by @javydekoning.
- Upgrade MegaLinter to v7 and enable cfn-lint (#632), by @javydekoning.
- Fix linter failures (#637) by @javydekoning.
- Linter fixes (#646) by @javydekoning.
- Add docs enhancement regarding ADF and AWS Control Tower (#638) by @AndyEfaa.
- Fix include all tests in pytest.ini for bootstrap CodeBuild project (#621) by @AndyEfaa.
- Remove CodeCommitRole from initial base stack (#663) by @alFReD-NSH.
- Fix bootstrap pipeline tests (#679) by @sbkok.
- Add AccessControl property on S3 Buckets (#681) by @sbkok.
- Version bump GitHub actions (#704) by @javydekoning, resolves #698.
- Bump express from 4.17.3 to 4.19.2 in /samples/sample-fargate-node-app (#697) by @dependabot.
- Update copyright statements and license info (#713) by @sbkok.
- Fix dead-link in docs (#707) by @javydekoning.
- Add BASH_SHFMT linter + linter fixes (#709) by @javydekoning.
- Fix sample expunge VPC, if-len, and process deployment maps (#716) by @sbkok.
- Moving CDK example app to latest CDK version (#706) by @javydekoning, resolves #618.
- Fix Markdown Anchor Link Check (#722) by @sbkok.
- Improve samples (#718) by @sbkok.
- Explain special purpose of adf-bootstrap/global.yml in docs (#730) by @sbkok, resolves #615.
- Rename `deployment_account_bucket` to `shared_modules_bucket` (#732) by @sbkok.
- Moved CodeCommit and EventBridge templates from lambda to the bootstrap repository to ease maintenance (#732) by @sbkok.

---


javydekoning (approved):

StewartW (approved):

Alexpngal (approved):
Nice work!

Issue #755, opened 2024-08-27
[Bug]:  adf-glabal-base-deployment stuck for missing permission in adf-bootstrap-update-deployment-role


giacomodxc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When adf-glabal-base-deployment stack ran it get stuck due to iam:PassRole missing permission.
<img width="539" alt="Screenshot 2024-08-27 090946" src="https://github.com/user-attachments/assets/da7f5d88-ca28-4a44-a3f7-cabb0aa02c57">

### Expected Behavior

The adf-glabal-base-deployment cloudformation stack should ran without error.

### Current Behavior

Any operation that will trigger the bootstrap pipeline which in turn ran the Cloudformation stack adf-global-base-deployment suffer from this behavior. i.e. changing account OU -

### Possible Solution

Probably we can consider adding the missing permission.

### ADF Version

4.0.0

### Contributing a fix?

- [ ] Yes, I am working on a fix to resolve this issue

PR #776, opened 2024-10-25
Fix deployment bootstrap IAM PassRole permissions

sbkok:
Issue: #755

## Why?

When an update is performed in the bootstrap repository, it will run `sam build` to generate the bootstrap stack for the deployment account.

This, however, includes new versions of some of its dependencies and therefore requires the Lambda Functions to update. While updating, it requires the `iam:PassRole` permission to pass the role to the new Lambda Function version.

This was not permitted by the update deployment role as used for minor updates. As reported in the issue linked above.

## What?

Updated the update deployment bootstrap role to include the required permissions to pass those roles as required.

Unfortunately, some of the Lambda functions relied on the `Policies` feature of SAM. This would auto generate a name for the role, thereby making it impossible to lock down permissions to the bare minimum. Hence, those functions now rely on dedicated Roles such that we can list the ARNs properly.

Half of the policies for the updated bootstrap deployment role have been relocated to an IAM Managed Policy to work around the 10k inline-policy limit.

Additionally, the permission to perform the `codebuild:BatchGetProjects` on the pipeline management CodeBuild project was missing.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/global.yml
@@ -1620,19 +1620,38 @@ Resources:
       Version: !Ref ADFVersion
       RepositoryArn: !GetAtt CodeCommitRepository.Arn
 
+  DetermineDefaultBranchFunctionRole:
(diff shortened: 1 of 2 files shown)


thomasmcgannon (approved):
LGTM

PR #786, opened 2025-02-24
fix the allow_empty_target OU path issue

danxie1999:
# Why?

When I enable the allow_empty_target parameter in the adfconfig.yml file, if I set a target OU path which does not exist in my AWS organization, I will get following error.

```
{
  "errorMessage": "Path /TenantXXX/Shared failed to return a child OU at 'TenantXXX'",
  "errorType": "ValueError",
  "requestId": "d786ed75-3ce9-4995-a7b6-75fa40454c86",
  "stackTrace": [
    "  File \"/var/task/generate_pipeline_inputs.py\", line 223, in lambda_handler\n    pipeline_input_data = generate_pipeline_inputs(\n",
    "  File \"/var/task/generate_pipeline_inputs.py\", line 151, in generate_pipeline_inputs\n    pipeline_target.fetch_accounts_for_target()\n",
    "  File \"/opt/python/target.py\", line 319, in fetch_accounts_for_target\n    self._target_is_ou_path(\n",
    "  File \"/opt/python/target.py\", line 260, in _target_is_ou_path\n    responses = self.organizations.get_accounts_in_path(\n",
    "  File \"/opt/python/organizations.py\", line 384, in get_accounts_in_path\n    ou_id = self.ou_path_to_id(path) if not ou_id else ou_id\n",
    "  File \"/opt/python/organizations.py\", line 367, in ou_path_to_id\n    raise ValueError(\n"
  ]
}
```

My deployment map file look like below:
```
pipelines:
  - name: platform-demo-empty-target
    description: |-
      Deploys Demo Stack to test ADF features
    default_providers:
      source:
        provider: codecommit
      build:
        provider: codebuild
    params:
      restart_execution_on_update: True
    targets:
      - name: Deploy-Non-Exist-Ou-Path
        path: /TenantXXX/Shared
        properties:
          template_filename: DemoStack1.template
```

## What?

The reason of above error is that the `src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py` only capture the boto3 Client Error and ignore the Error raised in the method get_accounts_in_path in orgnizations.py. So, an easy fix is to change the  boto3 Client Error to Exception and ignore it if the allow_empty_target was enabled.


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/target.py
```diff
@@ -275,11 +275,8 @@ def _target_is_ou_path(self, resolve_children=False):
                 ou_id=None,
                 excluded_paths=[],
             )
-        except ClientError as client_err:
-            no_target_found = (
-                client_err.response["Error"]["Code"] == "ParentNotFoundException"
-            )
-            if no_target_found and self.adf_deployment_maps_allow_empty_target:
+        except Exception:
+            if self.adf_deployment_maps_allow_empty_target:
```
sbkok, line 279: I don't think we should broaden the except type, as this would imply unrelated exceptions (think network or auth errors) would be hidden if `adf_deployment_maps_allow_empty_target` is set to true.

Could you please revert this change, then add support for another except that handles the error that it throws?

sbkok (changes requested):

PR #787, opened 2025-02-24
Feat/default codebuild image

danxie1999:
# Why?

Now in the deployment map, if the codebuild image does not explicitly defined in the deployment map, the deployment will fail. I would like to add define a default build image for the ADF. 

## What?
An easy fix is to define the default build image in the `src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py` and set the build stage to optional in the `src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/schema_validation.py`.

Additionally, `account_id` should also be optional in the schema validation, as ADF 4.0 supports default source account id.

### DIFF
--- a/docs/providers-guide.md
+++ b/docs/providers-guide.md
@@ -228,7 +228,7 @@ Provider type: `codebuild`.
 
 #### Properties
 
-- *image* *(String|Object)*.
+- *image* *(String|Object)*. - default: `STANDARD_5_0`
   - It is required to specify the container image your pipeline requires.
   - Specify the Image that the AWS CodeBuild will use. Images can be found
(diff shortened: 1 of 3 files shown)


sbkok (changes requested):
The CodeBuild Image was made a required property on purpose in ADF v4.0.
As setting a default image implies we need to release a major release to change it, as it would require introducing a breaking change.

Could you change this PR, or maybe better, open a new one for the CodeCommit account id optional change?

sbkok:
As linked in the message directly above, this PR is stale and superseded by #825. Hence closing this PR.

Issue #788, opened 2025-02-26
[Bug]: Organizations API throttling


igordust:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

In an installation with a fair amount of accounts (more than 250) and a pretty Organization structure, we're facing multiple issues regarding Organizations API throttling.

Most of the times, the throttling regard `ListParents` and `ListRoots` API calls 

These throttlings cause slowdowns or errors in the execution of the step functions adf-account-management and adf-account-bootstrapping, especially in the node EnableBootstrappingJumpRole. 

These are the output we observed on the EnableBootstrappingJumpRole node:
```json
{
  "error": "Task failed. Granting the ADF Account-Bootstrapping Jump Role privileged cross-account access failed due to an error: An error occurred (TooManyRequestsException) when calling the ListParents operation (reached max retries: 4): You have sent too many requests in too short a period of time. Try again later.."
}
```

```json
{
  "error": "Task failed. Granting the ADF Account-Bootstrapping Jump Role privileged cross-account access failed due to an error: An error occurred (TooManyRequestsException) when calling the ListRoots operation (reached max retries: 4): You have sent too many requests in too short a period of time. Try again later.."
}
```

### Expected Behavior

The step functions should complete without errors, and perform in a timely manner

### Current Behavior

Step functions randomly fail, take very long time to complete (~50 minutes)

### Possible Solution

Use some form of caching for Organizations API calls

### ADF Version

4.0.0

### Contributing a fix?

- [x] Yes, I am working on a fix to resolve this issue

Issue #784, opened 2025-02-17
[Bug]: adf-account-bootstrapping failing EnableBootstrappingJumpRole with ListParents too many requests


angelo-gelmini:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

ADF 4.0.0
More than 160 accounts, around 20 in protected ous.

If we bootstrap more than 1 account per time (creating or moving from protected is the same) at least State Machine "adf-account-bootstrapping" is failing running lambda "EnableBootstrappingJumpRole" with the following error:

_Task failed. Granting the ADF Account-Bootstrapping Jump Role privileged cross-account access failed due to an error: An error occurred (TooManyRequestsException) when calling the ListParents operation (reached max retries: 4): You have sent too many requests in too short a period of time. Try again later.._

The error is different from https://github.com/awslabs/aws-deployment-framework/issues/781 since the API call of the error is different.

### Expected Behavior

You can bootstrap more than one account per time.

### Current Behavior

You can not bootstrap more than one account per time.

The lambda stack trace is:
025-02-17 16:20:02,890 | ERROR | main | An error occurred (TooManyRequestsException) when calling the ListParents operation (reached max retries: 4): You have sent too many requests in too short a period of time. Try again later. | (main.py:464)
Traceback (most recent call last):
File "/var/task/main.py", line 499, in _handle_event
result = _process_update_request(
^^^^^^^^^^^^^^^^^^^^^^^^
File "/var/task/main.py", line 389, in _process_update_request
non_bootstrapped_account_ids = _get_non_bootstrapped_accounts(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/var/task/main.py", line 246, in _get_non_bootstrapped_accounts
_get_non_special_privileged_access_account_ids(
File "/var/task/main.py", line 204, in _get_non_special_privileged_access_account_ids
_get_non_special_adf_accessible_accounts(
File "/var/task/main.py", line 179, in _get_non_special_adf_accessible_accounts
adf_accessible_accounts = organizations.get_accounts(
^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/opt/python/organizations.py", line 296, in get_accounts
if self._account_available_to_adf(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/opt/python/organizations.py", line 252, in _account_available_to_adf
[...]

Issue #781, opened 2025-02-05
[Bug]: adf-account-bootstrapping SFN triggering exception TooManyRequestsException in EnableBootstrappingJumpRole


igordust:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the bug

Running the adf-account-bootstrapping SFN, sometimes we're hitting the exception TooManyRequestsException in the node EnableBootstrappingJumpRole, this causes the bootstrap to fail randomly.

### Expected Behavior

The Step Function should complete without issues

### Current Behavior

I'm working on a Landing Zone supporting about 250 accounts.
I observed in a couple of occasions this error below:
```
... assume role on each account ...
2025-02-04 15:39:19,087 | INFO | sts | Assumed into arn:aws:iam::111111111111:role/adf/bootstrap/adf-bootstrap-test-role with session name: jump_role_manager | (sts.py:42)
2025-02-04 15:39:19,252 | INFO | sts | Assumed into arn:aws:iam::222222222222:role/adf/bootstrap/adf-bootstrap-test-role with session name: jump_role_manager | (sts.py:42)
2025-02-04 15:39:31,287 | ERROR | main | An error occurred (TooManyRequestsException) when calling the ListRoots operation (reached max retries: 4): You have sent too many requests in too short a period of time. Try again later. | (main.py:464)
Traceback (most recent call last):
File ""/var/task/main.py"", line 499, in _handle_event
result = _process_update_request(
^^^^^^^^^^^^^^^^^^^^^^^^
File ""/var/task/main.py"", line 389, in _process_update_request
non_bootstrapped_account_ids = _get_non_bootstrapped_accounts(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File ""/var/task/main.py"", line 246, in _get_non_bootstrapped_accounts
_get_non_special_privileged_access_account_ids(
File ""/var/task/main.py"", line 209, in _get_non_special_privileged_access_account_ids
1738683571288,'"+ _get_filtered_non_special_root_ou_accounts(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File ""/var/task/main.py"", line 128, in _get_filtered_non_special_root_ou_accounts
organizations.get_ou_root_id(),
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File ""/opt/python/organizations.py"", line 354, in get_ou_root_id
[...]

PR #789, opened 2025-02-26
Feature/organization caching

igordust:
# Why?

With version 4.0, we noticed an increase in the number of Organizations API Calls that are performed during account management and bootstrap, leading to errors or longer execution times.

*Issue #788, #784, #781

## What?

Description of changes:

Now Organizations class always use the Cache object, even when it's not passed to the constructor. If passed in the constructor, allow the user of the class to share the same cache with different instances of Organizations. This happens for example in the bootstrap pipeline, where a worker thread for each account is created.
The following methods now make use of Cache:
* get_organization_info
* describe_ou_name
* describe_account_name
* list_parents
* get_ou_root_id
* list_organizational_units_for_parent
* build_account_path

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
@@ -256,14 +256,14 @@ def worker_thread(
 
     organizations = Organizations(
         role=boto3,
-        account_id=account_id
+        account_id=account_id,
+        cache=cache
     )
     ou_id = organizations.get_parent_info().get("ou_parent_id")
 
     account_path = organizations.build_account_path(
         ou_id,
-        [],  # Initial empty array to hold OU Path,
-        cache
+        [],  # Initial empty array to hold OU Path
     )
     try:
         role = ensure_generic_account_can_be_setup(
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
@@ -262,8 +262,9 @@ def worker_thread(
     ou_id = organizations.get_parent_info().get("ou_parent_id")
 
     account_path = organizations.build_account_path(
-        ou_id,
-        [],  # Initial empty array to hold OU Path
+        ou_id=ou_id,
+        account_path=[],  # Initial empty array to hold OU Path
+        cache=cache
     )
     try:
         role = ensure_generic_account_can_be_setup(
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
@@ -257,7 +257,7 @@ def worker_thread(
     organizations = Organizations(
         role=boto3,
         account_id=account_id,
-        cache=cache
+        cache=cache,
     )
     ou_id = organizations.get_parent_info().get("ou_parent_id")
 
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
@@ -264,7 +264,6 @@ def worker_thread(
     account_path = organizations.build_account_path(
         ou_id=ou_id,
         account_path=[],  # Initial empty array to hold OU Path
-        cache=cache,
     )
     try:
         role = ensure_generic_account_can_be_setup(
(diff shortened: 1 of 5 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -408,14 +426,14 @@ def build_account_path(self, ou_id, account_path, cache):
...
             # check cache for ou name of id
-            if not cache.exists(current.get("Id")):
-                cache.add(
+            if not self.cache.exists(current.get("Id")):
+                self.cache.add(
                     current.get("Id"),
                     self.describe_ou_name(current.get("Id")),
                 )
-            ou_name = cache.get(current.get("Id"))
+            ou_name = self.cache.get(current.get("Id"))
```
sbkok, line 434: We don't need to cache this at this layer anymore, as you added caching in the describe_ou_name function 👍 

Hence, this would suffice:
```suggestion
            ou_name = self.describe_ou_name(current.get("Id"))
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -408,14 +426,14 @@ def build_account_path(self, ou_id, account_path, cache):
         # While not at the root of the Organization
         while current.get("Type") != "ROOT":
             # check cache for ou name of id
-            if not cache.exists(current.get("Id")):
-                cache.add(
+            if not self.cache.exists(current.get("Id")):
+                self.cache.add(
                     current.get("Id"),
                     self.describe_ou_name(current.get("Id")),
                 )
```
sbkok, line 433: This is not required, see next comment.

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
```diff
@@ -440,16 +458,23 @@ def get_account_ids_for_tags(self, tags):
...
             )
             for ou in org_units["OrganizationalUnits"]
         ]
-        return organizational_units
+
+    def list_organizational_units_for_parent(self, parent_ou):
+        print(f'looking for children in {parent_ou}')
+        cache_key = f'children_{parent_ou}'
+        if not self.cache.exists(cache_key):
+            LOGGER.debug(f'Cache MISS for children of OU {parent_ou}')
```
sbkok, line 475:
```suggestion
            LOGGER.debug('Cache MISS for children of OU %s', parent_ou)
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
```diff
@@ -256,14 +256,14 @@ def worker_thread(
...
+        account_id=account_id,
+        cache=cache
     )
     ou_id = organizations.get_parent_info().get("ou_parent_id")
 
     account_path = organizations.build_account_path(
         ou_id,
-        [],  # Initial empty array to hold OU Path,
-        cache
+        [],  # Initial empty array to hold OU Path
```
sbkok, line 266: Could you also apply the same patch to the other usage of `build_account_path` on line 494 in this file?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
```diff
@@ -256,14 +256,15 @@ def worker_thread(
...
     )
     ou_id = organizations.get_parent_info().get("ou_parent_id")
 
     account_path = organizations.build_account_path(
-        ou_id,
-        [],  # Initial empty array to hold OU Path,
-        cache
+        ou_id=ou_id,
+        account_path=[],  # Initial empty array to hold OU Path
+        cache=cache,
```
sbkok, line 267:
```suggestion
```

Can you also remove the cache parameter on line 495?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py
sbkok: The following block can be removed from this function (unfortunately Github does not let me add this as a suggestion):

```
            if not self.cache.exists(current.get("Id")):
                self.cache.add(
                    current.get("Id"),
                    self.describe_ou_name(current.get("Id")),
                )
```

sbkok (changes requested):

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
sbkok: Could you please remove the cache parameter in the other build_account_path call on line 491?

sbkok (approved):
LGTM, thank you!

StewartW (approved):

PR #794, opened 2025-03-04
feat: add support for AWS China regions

danxie1999:
# Why?

Now the we cannot deploy the AWS Deployment Framework (ADF) in AWS China regions. This pull request introduces support for the AWS China Regions in ADF.

AWS China now have two regions which are cn-north-1 and cn-northwest-1. While the cn-north-1 is the primary region, the AWS Organization events only available in cn-northwest-1.

From this perspective, additional event forwarder is required in cn-northwest-1 (aws-deployment-framework/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/china-support). 

Additionally, In China region, the S3 URL is in following format: https://{self.bucket}.s3.{self.region}.{self.domain_suffix}/{key} (aws-deployment-framework/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/s3.py ).

## What?

Description of changes:
- In the codebuild buildspec files, add pip or npm China mirrors to fix the network issue.
- In the pytest, conformation files, change the "arn:aws" to "arn:{PARTITION}" to fix the partition issue.
- Deploy additional resources in the cn-northwest-1 to enable the Organizations events like "move OU" and forward to the primary region "cn-north-1".
- in the cloudformation.py, add additional parameters local_template_path and template_file_prefix to support local template path and customized template file path.

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/cloudformation.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/cloudformation.py
@@ -16,6 +16,8 @@ import tenacity
 from errors import InvalidTemplateError, GenericAccountConfigureError
 from logger import configure_logger
 from paginator import paginator
+from partition import get_partition
+
 
 LOGGER = configure_logger(__name__)
 STACK_TERMINATION_PROTECTION = os.environ.get('TERMINATION_PROTECTION', False)
@@ -29,7 +31,8 @@ CFN_CONFIG = Config(
 CFN_UNACCEPTED_CHARS = re.compile(r"[^-a-zA-Z0-9]")
 ADF_GLOBAL_IAM_STACK_NAME = 'adf-global-base-iam'
 ADF_GLOBAL_BOOTSTRAP_STACK_NAME = 'adf-global-base-bootstrap'
-
+ADF_GLOBAL_BOOTSTRAP_CHINA_BUCKET_STACK_NAME = "adf-regional-base-china-bucket"
+ADF_GLOBAL_BOOTSTRAP_CHINA_EXTRA_STACK_NAME = "adf-regional-base-china-extra"
 
 class StackProperties:
     clean_stack_status = [
@@ -105,6 +108,7 @@ class StackProperties:
             else None
         )
         self.s3 = s3
+        self.partition = get_partition(region)
         self.stack_name = stack_name or self._get_stack_name()
 
     def _get_geo_prefix(self):
@@ -147,6 +151,10 @@ class StackProperties:
         if self.region == self.deployment_account_region:
             valid_stack_names.append(ADF_GLOBAL_IAM_STACK_NAME)
             valid_stack_names.append(ADF_GLOBAL_BOOTSTRAP_STACK_NAME)
+
+        if self.partition == "aws-cn":
+            valid_stack_names.append(ADF_GLOBAL_BOOTSTRAP_CHINA_BUCKET_STACK_NAME)
+            valid_stack_names.append(ADF_GLOBAL_BOOTSTRAP_CHINA_EXTRA_STACK_NAME)
 
         return valid_stack_names
 
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/china-support/create_s3_cn.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/china-support/create_s3_cn.py
@@ -14,14 +14,14 @@ from cloudformation import CloudFormation
 REGION_DEFAULT = os.environ["AWS_REGION"]
 MANAGEMENT_ACCOUNT_ID = os.environ["MANAGEMENT_ACCOUNT_ID"]
 LOGGER = configure_logger(__name__)
-
+ADF_GLOBAL_BOOTSTRAP_CHINA_BUCKET_STACK_NAME = "adf-regional-base-china-bucket"
 
 def _create_s3_bucket(bucket_name):
     try:
         LOGGER.info("Deploy S3 bucket %s...", bucket_name)
         extra_deploy_region = "cn-northwest-1"
         template_path = "adf-build/china-support/cn_northwest_bucket.yml"
-        stack_name = "adf-regional-base-china-bucket"
+        stack_name = ADF_GLOBAL_BOOTSTRAP_CHINA_BUCKET_STACK_NAME
         parameters = [
             {
                 "ParameterKey": "BucketName",


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/main.py
```diff
@@ -459,6 +465,49 @@ def _sfn_execution_exists_with(
...
+        except Exception as error:
+            LOGGER.error(
+                "China extra stack adf-regional-base-china-extra deployment failed in region "
+                "%(region)s, please check following error: %(error)s",
+                {
+                    "region": CHINA_SECONDARY_REGION,
+                    "error": str(error),
+                },
+            )
+            sys.exit(2)
```
sbkok, line 510: Moving the if statement to the location where we call this method.

```suggestion
def _china_region_extra_deploy():
    parameters = [
        {
            'ParameterKey': 'AccountBootstrapingStateMachineArn',
            'ParameterValue': ACCOUNT_BOOTSTRAPPING_STATE_MACHINE_ARN,
            'UsePreviousValue': False,
        },
        {
            'ParameterKey': 'AdfLogLevel',
            'ParameterValue': ADF_LOG_LEVEL,
            'UsePreviousValue': False,
        },
    ]
    try:
        s3_china = S3(
            region=REGION_DEFAULT,
            bucket=S3_BUCKET_NAME
        )
        cloudformation = CloudFormation(
            region=CHINA_SECONDARY_REGION,
            deployment_account_region=CHINA_SECONDARY_REGION,
            role=boto3,
            wait=True,
            stack_name=ADF_REGIONAL_BASE_CHINA_EXTRA_STACK_NAME,
            s3=s3_china,
            s3_key_path='adf-build',
            account_id=MANAGEMENT_ACCOUNT_ID,
            template_file_prefix=CHINA_SECONDARY_REGION_DEPLOY_TEMP,
            parameters=parameters
        )
        cloudformation.create_stack()
    except Exception as error:
        LOGGER.error(
            "China extra stack adf-regional-base-china-extra deployment failed in region "
            "%(region)s, please check following error: %(error)s",
            {
                "region": CHINA_SECONDARY_REGION,
                "error": str(error),
            },
        )
        sys.exit(2)
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/cloudformation.py
```diff
@@ -189,6 +199,12 @@ def __init__(
             s3=s3,
             s3_key_path=s3_key_path
         )
+        self.template_url_from_template_file_prefix = self.s3.fetch_s3_url(
+            self._create_template_path(self.s3_key_path, template_file_prefix)
+        ) \
+            if template_file_prefix else None
```
sbkok, line 205:
```suggestion
        self.template_url_from_template_file_prefix = (
            self.s3.fetch_s3_url(
                self._create_template_path(self.s3_key_path, template_file_prefix),
            ) if template_file_prefix else None
        )
```

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/partition.py
```diff
@@ -10,39 +10,47 @@
...
+    return "us-east-1"
+
 
-    return 'us-east-1'
+def get_aws_domain(region_name: str) -> str:
+    """
+    Get AWS domain suffix
+    """
+    if region_name.startswith("cn-north"):
+        return f"amazonaws.com.{region_name.split('-')[0]}"
```
sbkok, line 55: If the region_name starts with cn-, the `region_name.split('-')[0]` would always be `cn`.
Hence we can simplify this to:

```suggestion
        return f"amazonaws.com.cn"
```

src/lambda_codebase/organization/main.py
```diff
@@ -72,14 +72,16 @@ def as_cfn_response(self) -> Tuple[PhysicalResourceId, Data]:
...
+        'us-gov-west-1',
+        'cn-north-1'
     ]
     region = os.getenv('AWS_REGION')
 
     if region not in approved_regions:
         raise ValueError(
             "Deployment of ADF is only available via the us-east-1 "
             "and us-gov-west-1 regions."
+            "and cn-north-1 regions."
```
sbkok, line 84:
```suggestion
            "Deployment of ADF is only supported in us-east-1, "
            "us-gov-west-1, or cn-north-1 regions."
```

src/template.yml
```diff
@@ -1568,6 +1576,59 @@ Resources:
...
+              - "iam:PassRole"
+              - "iam:CreatePolicy"
+              - "iam:CreateRole"
+              - "iam:DeleteRole"
+              - "iam:DeleteRolePolicy"
+              - "iam:GetRole"
+              - "iam:PutRolePolicy"
+              - "iam:UpdateAssumeRolePolicy"
+              - "iam:TagRole"
+              - "iam:DeleteRole"
```
sbkok, line 1599: This is too permissive, can you update this to use the JumpRoleManager instead?
Similar to how the non-cn partitions build with CodeBuild?

src/template.yml
```diff
@@ -1568,6 +1576,59 @@ Resources:
...
+
+            Resource:
+              - !Sub "arn:${AWS::Partition}:lambda:*:${AWS::AccountId}:function:ForwardStateMachineFunction"
+              - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/adf-china-extra/adf-regional-base-*"
+              - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/adf-regional-base-*"
+          - Effect: "Allow"
+            Action:
+              - "s3:*"
```
sbkok, line 1607: Please restrict this to the exact actions required

src/template.yml
```diff
@@ -1568,6 +1576,59 @@ Resources:
...
+            Resource:
+              - !Sub "arn:${AWS::Partition}:s3:::adf-china-bootstrap-cn-northwest-1-${AWS::AccountId}"
+              - !Sub "arn:${AWS::Partition}:s3:::adf-china-bootstrap-cn-northwest-1-${AWS::AccountId}/*"
+          - Effect: "Allow"
+            Action:
+              - "events:*"
```
sbkok, line 1613: Please limit to the event actions that CodeBuild needs.

src/template.yml
```diff
@@ -1568,6 +1576,59 @@ Resources:
...
+            Resource:
+              - !Sub "arn:${AWS::Partition}:events:cn-northwest-1:${AWS::AccountId}:rule/adf-regional-base-china*"
+          - Effect: "Allow"
+            Action:
+              - "cloudformation:*"
```
sbkok, line 1618: This is too permissive, can you restrict it further please?

src/template.yml
```diff
@@ -1601,9 +1662,7 @@ Resources:
                   - organizations:ListAccounts
                   - organizations:ListAccountsForParent
                   - organizations:DescribeAccount
-                  - organizations:ListOrganizationalUnitsForParent
-                  - organizations:ListRoots
-                  - organizations:ListChildren
+                  - organizations:List*
```
sbkok, line 1665: Instead of removing the specific actions required, can you add the action that was missing?

src/template.yml
```diff
@@ -1568,6 +1576,59 @@ Resources:
...
+              - "cloudformation:*"
+            Resource:
+              - !Sub "arn:${AWS::Partition}:cloudformation:cn-northwest-1:${AWS::AccountId}:stack/adf-regional-base-china-bucket/*"
+              - !Sub "arn:${AWS::Partition}:cloudformation:cn-northwest-1:${AWS::AccountId}:stack/adf-regional-base-china-extra/*"
+              - !Sub "arn:${AWS::Partition}:cloudformation:cn-northwest-1:aws:transform/Serverless-2016-10-31"
+          - Effect: "Allow"
+            Action:
+              - "cloudformation:ValidateTemplate"
+              - "cloudformation:List*"
+              - "cloudformation:Describe*"
```
sbkok, line 1627: Can you list the specific actions it needs please?

src/template.yml
```diff
@@ -2382,7 +2451,7 @@ Resources:
               - ssm:GetParameters
               - ssm:GetParameter
             Resource:
-              - !Sub "arn:${AWS::Partition}:ssm:*:${AWS::AccountId}:parameter/adf/*"
+              - !Sub "arn:${AWS::Partition}:ssm:*:${AWS::AccountId}:parameter/*"
```
sbkok, line 2454: This should not be changed. Please change the specific parameter you need to read to match the path it is granted access to instead.

```suggestion
              - !Sub "arn:${AWS::Partition}:ssm:*:${AWS::AccountId}:parameter/adf/*"
```

src/template.yml
```diff
@@ -2426,6 +2495,11 @@ Resources:
               - iam:UntagRole
             Resource:
               - !Sub "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/adf-update-cross-account-access-role"
+          - Effect: "Allow"
+            Action:
+              - iam:*
+            Resource:
+              - "*"
```
sbkok, line 2502: I assume this was added to debug something that didn't work.
We cannot add this, that would be way too permissive.

docs/admin-guide.md
```diff
@@ -964,8 +964,8 @@ To determine the current version, follow these steps:
 ### ADF version you have deployed
 
 To check the current version of ADF that you have deployed, go to the management
-account in us-east-1. Check the CloudFormation stack output or tag of the
-`serverlessrepo-aws-deployment-framework` Stack.
+account in us-east-1 or cn-north-1. Check the CloudFormation stack
```
sbkok, line 967: Instead of adding `cn-north-1` everywhere where `us-east-1` is used, I would like to suggest we add a mapping table at the top to explain what values to use for each partition.

Otherwise we should also add the us-gov region everywhere, in case that partition is used.
For the US-Gov partition, the region they should deploy to is `us-gov-west-1`.

docs/installation-guide.md
```diff
@@ -28,7 +28,7 @@ It is okay to install ADF and AWS Control Tower in different regions.
 For example:
 
 - Install AWS Control Tower in `eu-central-1`.
-- Install ADF in `us-east-1`.
+- Install ADF in `us-east-1` or `cn-north-1`.
```
sbkok, line 31: Does that work, having AWS Control Tower deployed in `eu-central-1` while deploying ADF in `cn-north-1`?

src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/china-support/tests/test_create_s3_cn.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Act
+    create_s3_cn._create_s3_bucket(bucket_name)
+
+    # Assert
+    # Remove this line as it's incorrect - we're not calling the mock directly
+    # mock_cloudformation.assert_called_once()
+
+    # Instead, check that create_stack was called on the mock instance
+    mock_cloudformation.create_stack.assert_called_once()
```
sbkok, line 61:
```suggestion
    # Assert
    mock_cloudformation.create_stack.assert_called_once()
```

sbkok (changes requested):
Thank you for contributing this feature.
Could you please look at the changes requested and accept/explain or improve those?
Thanks for your help on this.

PR #796, opened 2025-03-07
Feature disable tests

igordust:
# Why?

The bootstrap pipeline runs unit tests on ADF every time it is triggered. This implies the download of the development packages and the whole test suite to run, from my measurements, this adds 5-6 minutes to the overall execution time.

The current PR introduces a parameter in `adfconfig.yml`that disables the tests. If the parameter is not present, the tests are disabled by default.

## What?

Description of changes:

- Added lookup of the parameter `config.run-tests` in `adfconfig.yml`
- Split the python packages installation in runtime packages and dev packages
- put dev packages installation and test under the condition that `run-tests` = `enabled`
- Added the default parameter in adfconfig.yml.j2

---

### DIFF
--- a/docs/admin-guide.md
+++ b/docs/admin-guide.md
@@ -20,8 +20,7 @@
       - [Bootstrapping Recommendations](#bootstrapping-recommendations)
     - [Pipelines](#pipelines)
       - [Pipeline Parameters](#pipeline-parameters)
-      - [Using AWS CodeConnections for Bitbucket, GitHub, or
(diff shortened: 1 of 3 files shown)


sbkok (changes requested):
Thank you for contributing this optimization.
I agree, for most ADF users the code is not changed outside of deploying ADF updates.

However, the default should be enabled as that is the safest approach.

Could you also update the `docs/admin-guide.md` to include the new config option?

sbkok (dismissed):
LGTM, Thank you for contributing this.

sbkok (approved):

PR #797, opened 2025-03-20
fix(docs): Fix build status badges in readme

sbkok:
## Why?

GitHub changed the path for the badges.
Using the old path would show "No status" instead of the actual status.

Updated the path to fix it.

---


thomasmcgannon (approved):
LGTM

javydekoning (approved):

PR #815, opened 2026-07-01
fix(account-state): account "State" field over deprecated "Status"

### LINKED ISSUE #812: Deprecation of Status field in AWS Organizations API
AWS has announced the deprecation of the Status field in the AWS Organizations API in favor the more granular State field. The Status field will be permanently removed after September 9, 2026.

The [`organizations.py`](https://github.com/awslabs/aws-deployment-framework/blob/master/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/organizations.py) class in this project currently relies on the Status field, which will lead to failures or incorrect behavior once the deprecation is completed.

monicamico:
## Why?

AWS Organizations has introduced a new account `State` field that replaces the
legacy `Status` field, providing more granular account lifecycle information
(AWS Cloud Operations Blog, 15 Sep 2025).

- The legacy `Status` field exposes only three values: `ACTIVE`, `SUSPENDED`,
  `PENDING_CLOSURE`.
- The new `State` field exposes five: `PENDING_ACTIVATION`, `ACTIVE`,
  `SUSPENDED`, `PENDING_CLOSURE`, `CLOSED`.
- Both fields are returned by `DescribeAccount`, `ListAccounts`,
  `ListAccountsForParent`, and `ListDelegatedAdministrators` between
  September 2025 and September 2026.
- After 9 September 2026 the `Status` field is removed from API responses,
  leaving only `State`.

ADF filters accounts using `Status == "ACTIVE"` in several places. Once
`Status` is removed, `account.get("Status")` returns `None`, so every account
would be treated as inactive and skipped, breaking account bootstrapping,
account management, deployment-map target resolution, and the Terraform
helper org-wide.

Preferring `State` is also more correct today: an account can report
`Status == "ACTIVE"` while still completing sign-up, whereas `State`
distinguishes `PENDING_ACTIVATION` from `ACTIVE`. Filtering on
`State == "ACTIVE"` therefore avoids treating not-yet-activated accounts as
ready.

Reference:
https://aws.amazon.com/blogs/mt/updates-to-account-status-information-in-aws-organizations/

*Issue:* Resolve: #812

## What?

Description of changes:

- Rely on the new `State` field everywhere ADF filters accounts by activity:
  - adf-build/shared/python/organizations.py
  - adf-build/shared/python/target.py
  - adf-build/shared/helpers/retrieve_organization_accounts.py
  - adf-build/shared/helpers/terraform/get_accounts.py
  - lambda_codebase/account/main.py
- Add unit tests covering `State`-based filtering (including `PENDING_CLOSURE`
  and `SUSPENDED` exclusion).
- In `retrieve_organization_accounts.py`, return the `Status` as "ACTIVE",
  but only when the `State` is "ACTIVE", as all other accounts are filtered
  out anyways. So consumers requesting `Status` keep working now that the
  API no longer returns it standalone. Document the `State`/`Status` fields
  in the `--field` help text. This can be deprecated in the next major release.
- Add helper test coverage: `tests/test_retrieve_organization_accounts.py`
  (state->status ACTIVE mapping, active-only filtering, field selection,
  pagination, STS credential handling, output) and
  `tests/terraform/test_get_accounts.py` (active filtering, OU traversal,
  recursion, main), with a `tests/terraform` package `__init__.py`.
- Fix `helpers/pytest.ini` `norecursedirs` to `*/helpers/terraform` so the
  runnable `terraform` source dir stays excluded while the new
  `tests/terraform` tests are collected.

---

### DIFF
--- a/src/lambda_codebase/account/main.py
+++ b/src/lambda_codebase/account/main.py
@@ -151,7 +151,11 @@ def _find_deployment_account_via_orgs_api() -> str:
             DEPLOYMENT_OU_PATH,
         )
         active_accounts = list(filter(
-            lambda account: account.get("Status") == "ACTIVE",
+            # Prefer `State`, fall back to the deprecated `Status` field
+            # (removed after September 2026) for backward compatibility.
+            lambda account: (
+                account.get("State", account.get("Status")) == "ACTIVE"
+            ),
             accounts_found,
         ))
         number_of_deployment_accounts = len(active_accounts)
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/retrieve_organization_accounts.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/helpers/retrieve_organization_accounts.py
@@ -212,7 +212,12 @@ def _get_member_accounts(billing_account_id, options):
         accounts.extend(page["Accounts"])
 
     # Remove any account that is not actively part of this organization yet.
-    only_active_accounts = filter(lambda a: a["Status"] == "ACTIVE", accounts)
+    # Prefer the `State` field, falling back to the deprecated `Status` field
+    # (removed after September 2026) for backward compatibility.
+    only_active_accounts = filter(
+        lambda a: a.get("State", a.get("Status")) == "ACTIVE",
+        accounts,
+    )
 
     # Only return the key: value pairs that are defined in the --field option.
     only_certain_fields_of_active = list(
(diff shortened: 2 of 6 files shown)


sbkok:
Hi @moniami, thanks for driving these changes.
I've taken a look at them and need to modify the approach slightly. I will propose the required changes in this PR, so it still credits you for driving this. Currently, the "State" field is not recognized by the specific version of Boto that was used by ADF. Hence, I've created #816, to update the dependencies. Once that is merged, I will rebase this MR and apply another commit to get it ready for the next release of ADF. Thanks for your patience.

sbkok:
All dependencies have been updated. The version of boto used by ADF supports reading the State property.

PR #823, opened 2026-09-07
chore(ci): Stop GitHub Workflows running twice per commit

sbkok:
## Why?

The `push` trigger was unrestricted, so ADF and MegaLinter ran on every commit to every branch. When that branch also had an open pull request, the `pull_request` trigger fired for the same commit, producing two redundant runs.

## What?

Restrict the `push` trigger to the `main` and `master` branches for both CI workflows. Feature branch commits are now linted only via `pull_request`, while pushes to the mainline branches still run, eliminating the duplicate execution.

---

### DIFF
--- a/.github/workflows/adf.yml
+++ b/.github/workflows/adf.yml
@@ -1,6 +1,14 @@
 name: ADF CI
 
-on: [push, pull_request]
+on:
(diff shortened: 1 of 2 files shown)


thomasmcgannon (approved):
LGTM

PR #824, opened 2026-09-09
chore(ci): make lychee resilient to timeouts and exclude gnu.org

sbkok:
## Why?

The gnu.org URLs frequently lead to timeouts or rate-limited responses, thereby failing the workflow.

## What?

Add max_retries, retry_wait_time, and timeout so lychee retries transient failures (e.g. request timeouts) before failing the SPELL_LYCHEE step. Also exclude www.gnu.org/software/ links, which rate-limit and time out from the CI runner, and generalise the 429 accept comment accordingly.

---

### DIFF
--- a/lychee.toml
+++ b/lychee.toml
@@ -7,10 +7,9 @@
 #   - 300..=399: redirects. Combined with max_redirects = 0 below, lychee does
 #     not follow redirects and treats a 3xx response as a valid link, so
 #     harmless 301/302 redirects are not reported.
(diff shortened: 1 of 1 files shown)


javydekoning (dismissed):
LGTM, but open to consider allowlisting https://www.gnu.org/*

sbkok:
Good point, thanks for that. Addressed this accordingly :)

javydekoning (approved):

PR #825, opened 2026-09-10
fix(deployment-map): Allow omitting CodeCommit account_id and build

sbkok:
## Why

Resolves: #787.

At the moment, the deployment map schema requires an `account_id` on a `codecommit` source and requires a top-level `build` provider block on every pipeline. The runtime and the documentation, however, already treat both as optional: when the CodeCommit `account_id` is omitted it defaults to the deployment account id, and a pipeline without an explicit `build` block falls back to the default CodeBuild provider. The schema was simply stricter than the code and the docs, so valid deployment maps were rejected during validation before the runtime ever got a chance to apply its defaults.

This salvages the schema-only portion of upstream PR #787 by danxie1999. The default-`CODEBUILD_IMAGE` feature from that PR is deliberately not included here; only the two schema relaxations are.

## What

- In `CODECOMMIT_SOURCE_PROPS`, wrap `account_id` in `Optional(...)` so a `codecommit` source validates without it. The S3 source `account_id` stays required, as an S3 source cannot fall back to a default account.
- In `PROVIDER_SCHEMA`, make the top-level `build` key `Optional(...)` so a pipeline validates without an explicit build block.

No runtime or CDK wiring changes are needed; both defaults were already handled downstream. Tests are added that validate a `codecommit` source without an `account_id` and a pipeline without a `build` block through the public `SchemaValidation` interface, alongside negative cases ensuring a malformed account id and an invalid build provider still fail.

---

### DIFF
--- a/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/tests/test_adf_codepipeline_generate.py
+++ b/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/tests/test_adf_codepipeline_generate.py
@@ -4,6 +4,7 @@
 # pylint: skip-file
 
 from mock import patch
+from copy import deepcopy
 from cdk_constructs.adf_codepipeline import Action
 from adf_codepipeline_test_constants import BASE_MAP_PARAMS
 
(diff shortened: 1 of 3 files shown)


src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/tests/test_adf_codepipeline_generate.py
```diff
@@ -40,3 +41,55 @@ def test_generates_without_input_and_output_artifacts(input_mock, output_mock, a
...
+        'arn:aws:iam::222222222222:role/adf-codecommit-role'
+    )
+
+
+@patch('cdk_constructs.adf_codepipeline._codepipeline.CfnPipeline.ActionDeclarationProperty')
+def test_source_account_id_uses_default_scm_codecommit_account_id_override(action_decl_mock):
+    action_decl_mock.side_effect = lambda **x: x
+    map_params = deepcopy(BASE_MAP_PARAMS)
+    del map_params['default_providers']['source']['properties']['account_id']
+    map_params['scm/default_scm_codecommit_account_id'] = '333333333333'
```
javydekoning, line 86: Is the `scm/` prefix correct here? Isn't that the ssm prefix?  

It's unprefixed here: https://github.com/awslabs/aws-deployment-framework/blob/7add0773585660343b10e3c916630b9899393b49/src/lambda_codebase/initial_commit/bootstrap_repository/adf-bootstrap/deployment/lambda_codebase/pipeline_management/generate_pipeline_inputs.py#L191

javydekoning:
If a build block is omitted, we generate a default one here correct?

https://github.com/awslabs/aws-deployment-framework/blob/7add0773585660343b10e3c916630b9899393b49/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/python/pipeline.py#L75-L79

Then it defaults to `enabled = True` here:

https://github.com/awslabs/aws-deployment-framework/blob/7add0773585660343b10e3c916630b9899393b49/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_stacks/adf_default_pipeline.py#L128-L137

If no default is found

https://github.com/awslabs/aws-deployment-framework/blob/7add0773585660343b10e3c916630b9899393b49/src/lambda_codebase/initial_commit/bootstrap_repository/adf-build/shared/cdk/cdk_constructs/adf_codebuild.py#L366-L369

We ultimately return a 
`ValueError("Required CodeBuild image property is not configured")` here.

Should we throw sooner if `build` is omitted AND there is no default build image configured?

PR #826, opened 2026-09-16
sample(python-venv): Add Python CDK sample with independent CDK versioning

sbkok:
## Why

ADF pins its own CDK library and CLI versions for its build tooling, which can clash with an application that needs a different (often older) CDK version. Until now there was no worked example showing how to decouple an application's CDK version from the version ADF requires. This sample demonstrates the pattern: isolating each concern in its own Python virtual environment within a single CodeBuild job, so an application can pin its own CDK library and CLI versions independently of ADF.

## What

Add a new `samples/sample-python-cdk-app` sample that deploys a single "Hello World" Lambda via the AWS CDK for Python, structured to show:

- Three isolated venvs in `buildspec.yml` - `py-cdk-venv` (the app, `aws-cdk-lib==2.266.0`), `py-adf-venv` (ADF tooling, `aws-cdk-lib~=2.268.0`) and `py-lambda-venv` (the Lambda's own deps) - plus a separately pinned CDK CLI (`aws-cdk@2.1140.0`); none of them interfere because each lives in its own environment.
- An asset-free synth: `cdk synth` produces a template with no CDK-staged assets (bootstrap version rule disabled) so ADF deploys it through a plain CloudFormation action, without CDK bootstrap in the target accounts.
- Lambda code delivered via ADF's own `upload:`/`resolve:` parameter resolvers (`params/global_us-east-1.yml`), wiring the zipped package into the template's `AssetBucketParam` / `LambdaAssetObjectKeyParam` parameters.
- A README documenting the layout, the deployment-map entry and local synth.

---


javydekoning (dismissed):

sbkok:
@javydekoning While deploying the samples, I found two issues with this one. I've addressed those as separate commits, so it would not require a lot of time to review. A path issue with the SSM parameter lookup and the STANDARD_7_0 image usage in this, similar to #833.

javydekoning (approved):

PR #829, opened 2026-09-17
fix(installer): Update Python test for safe tarfile extraction

sbkok:
## Why

At the moment, the `verify_tooling` target compares only the major and
minor Python version (`3.9`) when deciding whether the interpreter is
recent enough. ADF's build relies on the PEP 706 safe tarfile extraction
filter, so any `3.9.0`-`3.9.16` passes this check up front, but the build
then fails later when a dependency invokes the filter, far from the root
cause.

Requiring a specific patch release cannot fix this in general either. The
backport landed in `3.9.17`, `3.10.12` and `3.11.4`, so `3.10.0`-`3.10.11`
and `3.11.0`-`3.11.3` are all `>= 3.9` yet still lack the filter. The
Python documentation recommends testing for the feature directly rather
than inferring it from the version number.

## What

- Replace the `REQUIRED_PYTHON_PATCH_VERSION` floor in `verify_tooling`
  with a direct `hasattr(tarfile, 'data_filter')` check, following the
  tarfile "Supporting older Python versions" guidance, while keeping the
  `3.9` minimum for the interpreter itself.
- Bump `MAKEFILE_VERSION` from `2.3.2` to `2.3.3`, since this changes the
  behaviour of the tooling check.

---


javydekoning:
Any patch version below 3.10.12 or 3.11.4 would still pass your min version check, but won't have the backport. 

Python docs state to test differently: https://docs.python.org/3.10/library/tarfile.html#supporting-older-python-versions

javydekoning:
I'd suggest reverting the patch version check and simplify to:

```python
sys.exit(
    sys.version_info < (3, 9)
    or not hasattr(tarfile, "data_filter")
)
```
As per official docs

sbkok:
Thanks for your review @javydekoning, great spot!
I've proposed a fix to address this the way you proposed.

javydekoning (approved):

